PR #4, opened 2015-12-23
uiSchema support

n1k0:
This ensures react-jsonschema-form uiSchema customization is supported by kinto-admin.

r=? @Natim


Natim:
r+

PR #14, opened 2016-03-09
Fixes #13: Upgrade react-jsonschema-form to 0.12

### LINKED ISSUE #13: Upgrade react-jsonschema-form to 0.12.0
Changelog: https://github.com/mozilla-services/react-jsonschema-form/releases/tag/v0.12.0

As we're starting from 0.4 currently, expect some breakage. /cc @magopian

n1k0:
r=? @magopian (let's pair review this tomorrow)


css/rjsf.css
```diff
@@ -1,25 +1,25 @@
-.generic-form * {
+.rjsf * {
```
magopian: is `rjsf` a fake/random name just for the tests, as I believe it is? If so, I'd rather have something like `random-name` or `test-name` or even `foobar` so it's explicit ;)

magopian:
a very small nit, other than that, r+

Thanks!

css/rjsf.css
n1k0: Nope, it's acronym for *r*eact *j*son *s*chema *f*orm and is the name of the provided CSS stylesheet root identifier for the built forms.

magopian: ah, makes total sense now that I see it, thanks for the explanation!

PR #25, opened 2016-03-22
Fixes #24 : CSS fix for the notification close button

### LINKED ISSUE #24: little notification "cross" is in position: fixed
This little cross (see screenshot below, its css class is `close`) is set as `position: fixed`, and will thus stay exactly in the same position when scrolling.

It's also not properly aligned with the notification box title.

 screen shot 2016-03-22 at 10 48 29

magopian:
Fixes #24

screen shot 2016-03-22 at 11 13 52


magopian:
@n1k0 @Natim what do you think?

PR #43, opened 2016-04-26
Fixes #42: Configurable settings from the json config file.

### LINKED ISSUE #42: Allow preconfiguring server settings in the admin config file

n1k0:
This patch updates the admin JSON configuration file to allow configuring default server settings. Note that this change breaks backward compatibility and require upgraders to update their config file. 

r=? @leplatrem @magopian

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,6 +58,7 @@
     "fake-indexeddb": "1.0.3",
     "file-loader": "^0.8.5",
     "gh-pages": "^0.11.0",
+    "history": "^1.17.0",
     "jsdom": "^8.3.0",
     "json-loader": "^0.5.4",
     "mocha": "^2.3.0",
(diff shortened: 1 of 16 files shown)


n1k0:
I need to ensure custom settings entered in the dedicated form are persisted locally; right now they're overridden by config settings every time the page is reloaded, which is wrong.

package.json
```diff
@@ -58,6 +58,7 @@
     "fake-indexeddb": "1.0.3",
     "file-loader": "^0.8.5",
     "gh-pages": "^0.11.0",
+    "history": "^1.17.0",
```
leplatrem: just curious, what is that ?

leplatrem:
LGTM :)

package.json
n1k0: A peer dependency which was handled automatically just fine with npm 2 but which breaks with npm 3; declaring it here makes it work in both versions :)

PR #47, opened 2016-04-27
use a pretty label for collection button

### LINKED ISSUE #40: Buckets and collections labels
*not an emergency*

Currently the object id is used as the label AFAIK.

The use-case would be to have a collection with id `addons` but shown as `Add-ons` in the menu.
It could read ~~a `label`~~ the `name` attribute in the collection metadata for example.

https://github.com/mozilla-services/kinto2xml/issues/13

glasserc:
Take the `name` field from the collection info in the config.

Fixes #40.

### DIFF
--- a/scripts/components/Sidebar.js
+++ b/scripts/components/Sidebar.js
@@ -29,9 +29,10 @@ export default class Sidebar extends Component {
                 "list-group-item",
                 params.name === name ? "active" : "",
               ].join(" ");
+              const label = collections[name].name || name;
               return (
                 <Link key={i} to={`/collections/${name}`} className={classes}>
-                  {name + (!collections[name].synced ? "*" : "")}
+                  {label + (!collections[name].synced ? "*" : "")}
                 </Link>
               );
             })
--- a/test/containers/Sidebar_test.js
+++ b/test/containers/Sidebar_test.js
@@ -18,7 +18,7 @@ describe("Sidebar container", () => {
     expect(nodeTexts(comp, "a")).eql([
       "Home",
       "Settings",
-      "tasks",
+      "Todo tasks",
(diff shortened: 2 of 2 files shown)


n1k0:
Looks great, thank you!

scripts/components/Sidebar.js
```diff
@@ -29,9 +29,10 @@ export default class Sidebar extends Component {
                 "list-group-item",
                 params.name === name ? "active" : "",
               ].join(" ");
+              const label = collections[name].name || name;
               return (
                 <Link key={i} to={`/collections/${name}`} className={classes}>
-                  {name + (!collections[name].synced ? "*" : "")}
+                  {label + (!collections[name].synced ? "*" : "")}
```
Natim: Does this still work with that change: https://github.com/mozilla-services/amo-blocklist-ui/commit/218b9bf8e4ebc300b5227eb5edf9e66fee8bf3bb

n1k0: Well `addons` will now be listed in the sidebar instead of "add-ons" if that's your question :) So to summarize:

- The collection key is the technical name (the Kinto collection name used in the request URLs);
- The name propery is the *display name* for that collection in the menu.

Natim: So I should have: 

```
{
-   "add-ons": {
+   "addons": {
     "name": "add-ons",
```

n1k0: Ah! If add-ons is what you want to render in the sidebar, yes :)

Natim: Addressed here: https://github.com/mozilla-services/amo-blocklist-ui/pull/95#issuecomment-215425791

Issue #59, opened 2016-05-09
Detect when the conflict actually contains twice the same record.


Natim:
In this specific case, both records contains the same data (even if it is hard to tell)

Picking one or the other will have the same impact.

![canonical-json](https://cloud.githubusercontent.com/assets/229453/15107350/9870dafe-15cf-11e6-864d-8fd003ddb464.png)

n1k0:
I'm very surprised kinto.js marks this as a conflict as it's not supposed to happen if both data objects are actually equals...

n1k0:
Also sounds like someything to do is process the json string representation to at least order the keys, so visually comparing these would be *much* easier.

Natim:
Apparently the schema value is different

n1k0:
That explains it.

Side note, you may be interested in tools like [json-diff](https://www.npmjs.com/package/json-diff) to quickly check for diffs. Another side note, we should probably leverage such a lib in the admin itself so we could display a diff!

n1k0:
Also, should the schema version value being different raise a conflict here? I feel like it shouldn't, as the two record data objects are compatible.

Natim:
Yes IMHO the schema value should be cleaned as for last_modified and id values.

leplatrem:
Closing since no offline sync anymore.

Issue #58, opened 2016-05-09
Display canonical JSON form to ease comparison


Natim:
![canonical-json](https://cloud.githubusercontent.com/assets/229453/15107307/5a5bf9f6-15cf-11e6-9ee8-9a80b027d69e.png)

Please tell me the differences between the two :)

n1k0:
last_modified differs.

n1k0:
Oh, the issue is about rendering a diff, okay ;)

n1k0:
Most interesting lib found for now is https://github.com/flitbit/diff

n1k0:
Fixed by #61

PR #61, opened 2016-05-10
Fixes #60: Improved conflict resolution UI.

### LINKED ISSUE #60: Render a diff for conflicting objects.

n1k0:
Refs #58, #60.

This attempts at iimproving the conflict resolution UI:

Note that this also addresses this comment https://github.com/Kinto/kinto-admin/issues/59#issuecomment-217804899.

r=? @glasserc @leplatrem @Natim

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,8 +24,10 @@
   },
   "dependencies": {
     "btoa": "^1.1.2",
+    "diff": "^2.2.2",
     "express": "^4.13.3",
     "extract-text-webpack-plugin": "^1.0.1",
+    "json-stable-stringify": "^1.0.1",
     "jsonschema": "^1.0.2",
     "kinto": "^2.0.4",
     "node-fs-extra": "^0.8.1",
--- a/scripts/components/Resolve.js
+++ b/scripts/components/Resolve.js
@@ -43,12 +43,12 @@ class VersionPicker extends Component {
         }</div>
         <div className="panel-footer text-center">
           <div className="btn-group">
-            <button type="button" className="btn btn-success"
+            <button type="button" className="btn btn-success btn-pick"
               onClick={() => pick(record)}>Pick {type} version</button>
             {diff ?
-              <button className="btn btn-info"
+              <button className="btn btn-info btn-diff"
                 onClick={this.hideDiff}>Hide diff</button> :
-              <button className="btn btn-info"
+              <button className="btn btn-info btn-diff"
                 onClick={this.showDiff}>Show diff</button>}
           </div>
         </div>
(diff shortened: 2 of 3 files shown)


glasserc:
I think you should reverse the directions of source and target. I don't have a rigorous justification, but on the left, I'd expect to see green corresponding to the local version, and I'd expect to see the remote version above the local version that "replaces" it.

glasserc:
I'm also not crazy about the diff being "chunks" instead of individual properties.

n1k0:
> I'm also not crazy about the diff being "chunks" instead of individual properties.

Yeah, I agree... I couldn't find any lib doing property-based diffs on npm, or the ones I found trying to were.. let's say they weren't really up to my expectations. We could probably spend some time writing our own, but that might get tricky when formatted arrays and deep nesting are involved.

If we could live with the current chunk-based diff approach, that'd probably save us some useful CPU cycles :) We could also spawn a new enhancement issue about it so we could land this (when it's ready) then iterate.

n1k0:
This is now ready for final review. r=? @glasserc @leplatrem @Natim

Natim:
If it is not possible to have red line on the left and green lines on the right, Is it possible to remove the diff altogether (red and green lines)?

n1k0:
> Is it possible to remove the diff altogether (red and green lines)?

Sorry, I didn't undersand

Natim:
Something like that: 
diff

Or if it is not possible the same thing without the + and colors.

n1k0:
I still don't get it and am finding the mockup even more confusing, but that may just be because I'm hands deep into the current implementation and so I got used to it :D

Maybe what you're actually suggesting is to have a side by side diff view ala Eclipse or equivalent?

Sorry I'm really trying to understand :)

Natim:
Yes basically my point is to have both records (the local one and the remote one) side by side with their properties sorted the same way so that it is easy to select one or the other.

If we can highlight properties that are different it is better but my point is to have the record in its entirety without mixed diff in it.

n1k0:
> Yes basically my point is to have both records (the local one and the remote one) side by side with their properties sorted the same way so that it is easy to select one or the other.

We already have that in the default view, did you try the patch ? :)

> If we can highlight properties that are different

That's the purpose of the diff view, but I understand you'd like to have them highlighted in the default view. That's probably achievable for simple one-line properties but I fear it can get tricky with formatted arrays. I'll try and see how it goes though.

Natim:
Oh I missed the button show diff :+1:  on the screenshot.let me try the patch :)

Natim:
r+

PR #65, opened 2016-05-18
Auth based sessions

n1k0:
This is a full rewrite of the kinto admin. I know.

It's now fully based on [kinto-client](https://github.com/Kinto/kinto-client), and doesn't do local persistence/sync at all anymore. This will ease a lot of things I'll write down here when I'll have the courage to. 

For now this is still WiP, but you can start looking at it with fear. i'm sorry :D

The new version of the kinto admin matching this patch has been deployed to [gh-pages](http://kinto.github.io/kinto-admin/) to ease the UX/QA review process.

#### Todo

- [x] Update `devServer.js`, drop any reference to the now obsolete json config file 
- [x] Clean `package.json` obsolete dependencies 
- [x] Missing some action tests
- [x] `kinto-admin` executable `build` & `serve` command
- [x] Update `package.json` scripts
- [x] Update docs

#### @glasserc feedback checklist

- [x] Collection page titles for default bucket shouldn't render the uuid but *default* instead
- [x] We should ensure urls load and put the admin in the expected state
- [ ] We should find a way to always render the uuid in the records list 

#### How to test this thing?

```
$ git checkout auth-based-sessions
$ npm install
$ npm start
```


glasserc:
@n1k0 suggested on IRC that he was looking for general feedback on the general features offered by the new patch, so I haven't really looked at the code yet. Since it's almost a complete a rewrite (3200 lines out of ~3600), I think I will have to look at the entire codebase from scratch, but I think this branch isn't quite ready for that yet.

Hitting reload doesn't work properly. I have to click "home" and then remove the hash part of the URL. Same with ctrl-clicking links to open in a new tab.

I tried to click "default bucket" at first before I realized it was a heading rather than a link. Also, I guess there's no way to administer buckets through the API, right? It seems silly to have only the default bucket if you can't get an overview of it, look at its settings or create other ones.

I don't like seeing only the JSON representation of the records instead of a tabular sort of representation based on its schema.

I'm not convinced "bulk add" provides enough value for whatever complexity it adds (although I don't know how much complexity that is). It seems like you would end up clicking just as many times to add N objects as you would if you clicked "add" N times.

I think the "back" button should be in the right column instead of the left. In the left column, it makes the sidebar jump around as you click on different records, and I tend to look for it just under the title of the page (like breadcrumbs).

It might be better to show the bucket's name instead of its ID. (I'm not sure buckets have names, but "e4736f03-61fd-a4a2-7734-89fe98742acb" is pretty unwieldy.)

There's no obvious way to get the ID of a single record (although I guess you can copy it out of the URL).

I think "server information" probably shouldn't be the front page of the admin after you log in, although there should be a page that displays it.

n1k0:
> I think this branch isn't quite ready for that yet.

I think it's actually quite ready. Read on.

> Hitting reload doesn't work properly. 

Same with current master.

> I have to click "home" and then remove the hash part of the URL. 

Same with current master.

> Same with ctrl-clicking links to open in a new tab.

Same with current master.

> Also, I guess there's no way to administer buckets through the API, right? 

There's an API but this patch doesn't leverage it yet. Do you really want a 6000 LOC diff?

> It seems silly to have only the default bucket if you can't get an overview of it, look at its settings or create other ones.

The dev server wired to the demo on gh-pages uses is flushed everyday and all existing buckets are deleted. If you create buckets using the API (curl style), You will see them listed in the UI, and you will be able to create new collections within them.

> I don't like seeing only the JSON representation of the records instead of a tabular sort of representation based on its schema.

How would you infer column headings out of a database having no schema definition? Don't forget a schema isn't mandatory in Kinto. So that means people may actually enter any kind of data, eg. a different data structure for each record entry.

> I'm not convinced "bulk add" provides enough value for whatever complexity it adds (although I don't know how much complexity that is). It seems like you would end up clicking just as many times to add N objects as you would if you clicked "add" N times.

Can you think of a better way to solve the need of being able to enter several records at once?

> I think the "back" button should be in the right column instead of the left. In the left column, it makes the sidebar jump around as you click on different records, and I tend to look for it just under the title of the page (like breadcrumbs).

Noted.

> It might be better to show the bucket's name instead of its ID. (I'm not sure buckets have names, but "e4736f03-61fd-a4a2-7734-89fe98742acb" is pretty unwieldy.)

Noted.

> There's no obvious way to get the ID of a single record (although I guess you can copy it out of the URL).

Where would you put it in the table? It's quite a long string.

> I think "server information" probably shouldn't be the front page of the admin after you log in, although there should be a page that displays it.

What would you display on the front page instead?

n1k0:
Oh, forgot to ask; is there any good in this new admin or is it as bad as it sounds? We can also decide to discard this initiative entirely, I'm easy with letting it go.

glasserc:
Oh, no, I actually think this direction makes a lot of sense. Moving to synchronous operation is definitely the way forward for kinto-admin, and once you accept that, it's practically a complete rewrite to get to that point. The only thing I actually care about is that we have an "official" demo of offline-first operation, but that doesn't have to be this. This version of kinto-admin feels a lot lighter and simpler because of the absence of synchronization and conflicts. Sorry, I didn't think to say that earlier.

> I think it's actually quite ready. Read on.

I'll look at the code as soon as I get the chance.

> Same with current master.

Oops! I guess I never tried that with current master, maybe because it doesn't feel as "light"?

> There's an API but this patch doesn't leverage it yet. Do you really want a 6000 LOC diff?

No, and that's a good point. Most of the things on my wishlist for this iteration are pretty complex and maybe we should leave them for another time.

> You will see them listed in the UI, and you will be able to create new collections within them.

Yes, I was just thinking that it if there's only one bucket and you can't create others, maybe we should display that differently, for example by dropping the header.

> Note that as soon a JSON schema is defined for a collection, columns will be shown according to the displayFields setting.

Ah! I didn't try that. That's pretty good. I assumed that was only useful if you wanted to hide certain fields, so didn't even think about using it. I would suggest that if a schema is defined but has no displayFields, we could use the schema itself to "guess" at display fields. That would be enough for me personally. We could also do what the AWS DynamoDB console does, and just create a table with the set of fields that is the union of the fields of the objects we're displaying at the moment. (And also assume that all fields are scalar, even though they provide lists and maps as types of fields.)

> Can you think of a better way to solve the need of being able to enter several records at once?

Maybe we could show 10 empty fields by default instead of 2, and assume that any object with completely empty fields should be silently dropped. 

> Where would you put it in the table? It's quite a long string.

I expected it to be shown on the edit page, or on the view page if there were a view page. It would be even nicer if there was a button you could click to copy it to the clipboard, but that's probably outside the scope of this PR. I don't think we need to show the ID but it's nice to be able to copy it if you want to share it with someone else.

> What would you display on the front page instead?

Probably something very similar to the current sidebar but bigger. (But the more I think about it, the less I'm sure that it's a good idea.)

leplatrem:
I just tried it locally, it makes a lot more sense than the previous sync indeed! That brings a lot of value to the ecosystem

Here are some ideas, feel free to ignore or spawn issues to avoid blocking the merge:

* In the collection create form, «Add» is bigger than «Submit». I was tempted to click «add», which adds a display field instead of the submit button on the left. — idea: replace «Add» by «+» and move submit to the right
* When going to «Create a collection», the item in the left menu is not blue like with other pages
* The «stack» icon in the left menu is not clickable, only the collection name label can be click to acces the list of records
* Create record / Bulk create buttons are not styled the same way as others
* Bulk form does not show field titles (eg. *JSON record*)
* The list with no json schema shows a single column with `<Unknown>`
* The last modified column in the list is truncated even though my screen is wide

* Add gear to open bucket object edition
* :bulb: add small external link close to «Json schema» in collection form that opens the form builder in a new tab
* :bulb:  being able to enter a bucket and collection name (or via the URL hash), because some user maybe be allowed to read some records without being authorized to read their parent bucket/collection

I made some efforts to find all those nits hey! None of those should block the merge anyway :)

scripts/actions/client.js
```diff
@@ -0,0 +1,207 @@
...
+
+export function listBuckets() {
+  return execute((client, dispatch, getState) => {
+    return client.fetchServerInfo()
+      .then((serverInfo) => {
+        dispatch(SessionActions.serverInfoLoaded(serverInfo));
+        // XXX We need to first issue a request to the default bucket in order
+        // to create it so we can retrieve the list of buckets.
+        // ref https://github.com/Kinto/kinto/issues/454
+        return client.bucket(serverInfo.user.bucket).getAttributes();
```
leplatrem: Be careful, if the `default_bucket` plugin is not installed, there won't be any `user.bucket`. You could inspect the capabilities or add a simple safety check

scripts/components/Sidebar.js
```diff
@@ -1,43 +1,90 @@
...
-                  {label + (!collections[name].synced ? "*" : "")}
-                </Link>
-              );
-            })
-          }</div>
-        </div>
+        {session.authenticated ?
+          <BucketsMenu
+            buckets={buckets}
+            userBucket={session.serverInfo.user.bucket}
```
leplatrem: This will be `undefined` if no default plugin. From what I read below, it won't pose any problem.

scripts/actions/client.js
```diff
@@ -0,0 +1,207 @@
...
+export let client;
+
+function getClient({session: {server, username, password}}) {
+  return client || new KintoClient(server, {
+    headers: {
+      Authorization: "Basic " + btoa([username, password].join(":")),
+      // XXX by default we must avoid caching authentication-dependent resources
+      // as they could leak private information to different users. Here we
+      // simply discard caching entirely, waiting for a Kinto server fix.
+      // ref https://github.com/Kinto/kinto/issues/593
```
leplatrem: Fixed with cliquet 3.1.5 https://github.com/mozilla-services/cliquet/releases/tag/3.1.5

scripts/actions/client.js
```diff
@@ -0,0 +1,207 @@
...
+      .then(({data}) => {
+        dispatch(notifySuccess("Collection deleted."));
+        dispatch(CollectionActions.collectionDeleted(data));
+        dispatch(listBuckets());
+        dispatch(updatePath(""));
+      });
+  });
+}
+
+export function loadCollectionProperties(bid, cid) {
```
leplatrem: nitpick: we use different wording for the same notion I believe (data, properties, attributes)

scripts/actions/client.js
```diff
@@ -0,0 +1,207 @@
...
+      .then(({data}) => {
+        dispatch(CollectionActions.collectionPropertiesLoaded({
+          ...data,
+          bucket: bid
+        }));
+      });
+  });
+}
+
+export function updateCollectionProperties(bid, cid, {schema, uiSchema, displayFields}) {
```
leplatrem: nitpick: I would leave `collectionData` in the signature and destructure in the body (mirror `createCollection()` function

leplatrem: rename to `updateCollection()` ?

scripts/actions/client.js
```diff
@@ -0,0 +1,207 @@
...
+    }, {aggregate: true})
+      .then((res) => {
+        if (res.errors.length > 0) {
+          const err = new Error("Some records could not be created.");
+          err.details = res.errors.map(err => err.error.message);
+          throw err;
+        } else {
+          dispatch(listRecords(bid, cid));
+          dispatch(updatePath(`/buckets/${bid}/collections/${cid}`));
+          dispatch(notifySuccess("Records created."));
```
leplatrem: nit: N records created

leplatrem:
awesome

scripts/actions/client.js
n1k0: There's a useful distinction; the `bid` form means that's a value we've received as a URL parameters, while `bucket` means this is something we received from the server.

n1k0:
Okay, I'm gonna land this and spawn new issues from the comments the patch didn't address.

n1k0:
Filed #66, #67, #68, #69, #70 and #71. If something's missing, please file an issue.

PR #77, opened 2016-05-21
Migrating action creators to Redux saga

n1k0:
This patch leverages [redux-saga](https://github.com/yelouafi/redux-saga/) and moved the action creators logic to dedicated sagas. Actions are now just simple synchronous descriptors, and sagas handle the asynchronous request stuff, triggering actions to notify the reducers when they're processed.

That makes the codebase better organized and easier to read especially because of the use of generators.
 
While the diff is large and seems to add more code than before, this is mostly because we now have much  more granular tests for action creators.

The patch has been deployed to [gh-pages](kinto.github.io/kinto-admin/).

#### Todo

- [x] Write missing saga tests
- [x] Figure out why we need *babel-polyfill* even for browsers that support generators natively


scripts/components/CollectionEdit.js
```diff
@@ -21,7 +21,7 @@ export default class CollectionEdit extends Component {
 
   render() {
     const {collection} = this.props;
-    const {label, schema, uiSchema, displayFields, busy} = collection;
+    const {schema={}, uiSchema={}, displayFields=[], label, busy} = collection;
```
leplatrem: wao, I didn't know about that syntax :) neat !

n1k0: Me neither 2 days ago :p Now I see so many places this should be used it's embarrassing.

n1k0:
This is now ready for, *ahem*, review. I'm pretty sure no one on Earth will ever want to thoroughly review this, so I'm gonna let it sit for a few hours and land it if I don't get any feedback meanwhile.

Pinging @Natim @leplatrem @glasserc @almet

Natim:
I will have a look at it :)

scripts/client.js
```diff
@@ -0,0 +1,25 @@
+import KintoClient from "kinto-client";
+
+
+let client;
+
+export function setupClient({server, username, password}) {
```
Natim: I wonder if we shouldn't pass the authorization header directly to later supports other authorization methods.

scripts/client.js
```diff
@@ -0,0 +1,25 @@
+import KintoClient from "kinto-client";
+
+
+let client;
```
Natim: I wonder if we shouldn't use a class that we can instanciate rather than a sigleton like that.

scripts/client.js
n1k0: yeah, but please let's iterate here :)

Natim: Ok :+1:

scripts/client.js
n1k0: I agree 100% and this is my main gripe with the patch and the previous implementation where it was the same.

The only way I could see having a state-dependent construct of the client would be to store it in the store, which sounds odd. Do you have an idea how we could avoid relying on a singleton in the current architecture? Or do you have another architecture in mind where achieving this would be easy?

Last, could we iterate here? :D

scripts/sagas/bucket.js
```diff
@@ -0,0 +1,111 @@
...
+  while(true) { // eslint-disable-line
+    const {bid, cid} = yield take(COLLECTION_LOAD_REQUEST);
+    yield fork(loadCollection, bid, cid);
+  }
+}
+
+export function* watchCollectionCreate() {
+  while(true) { // eslint-disable-line
+    const {bid, collectionData} = yield take(COLLECTION_CREATE_REQUEST);
+    yield fork(createCollection, bid, collectionData);
```
Natim: fork is a bit cryptic to me what do it do?

Natim: Apparently it is related to redux-saga I will probably have to read its doc :)

scripts/sagas/collection.js
```diff
@@ -0,0 +1,165 @@
...
+  }
+}
+
+export function* createRecord(bid, cid, record) {
+  try {
+    const coll = getCollection(bid, cid);
+    yield put(collectionActions.collectionBusy(true));
+    yield call([coll, coll.createRecord], record);
+    yield put(collectionActions.listRecords(bid, cid));
+    yield put(updatePath(`/buckets/${bid}/collections/${cid}`));
```
Natim: :+1:

scripts/sagas/session.js
```diff
@@ -0,0 +1,89 @@
...
+
+export function* listBuckets() {
+  const client = getClient();
+  try {
+    // XXX We need to first issue a request to the "default" bucket in order
+    // to create user associated permissions, so we can access the list of
+    // buckets.
+    // ref https://github.com/Kinto/kinto/issues/454
+    const buck = client.bucket("default");
+    yield call([buck, buck.getAttributes]);
```
Natim: We shouldn't for the creation of the default bucket here. Actually this makes the kinto-admin fails with kinto server that doesn't support default buckets such as : https://djangocong.herokuapp.com/v1/

scripts/sagas/bucket.js
n1k0: It basically schedule the execution of the saga in a non-blocking fashion. From the [docs](https://yelouafi.github.io/redux-saga/docs/api/index.html#forkfn-args):

> The middleware doesn't suspend the Generator while waiting for the result of fn. Instead as soon as fn is invoked, the Generator resumes immediately.

Natim:
There is tests :+1:

Natim:
I read the code, it looks good to me.

scripts/sagas/session.js
n1k0: Interesting, indeed I should check the serverInfo from state to see if we should ping the default bucket here. The thing is, I have no idea how to query global store state from a saga, neither if it's a good practice :D

scripts/sagas/collection.js
n1k0: Yeah I fixed things along the way :D

scripts/client.js
Natim: I would rather create a stateless client that we can create from the store.
I would then pass the serverInfo from the state if I catched them already.

It don't know if it makes sense.

n1k0: We could also store client state in the store, and have a dedicated saga for acting on it. I'll spend some time playing around with these idea, thank you!

scripts/sagas/session.js
n1k0: Note that an approach as described in https://github.com/Kinto/kinto-admin/pull/77#discussion_r64196279 would nicely solve this.

n1k0: Meanwhile added a check for default bucket information from serverInfo https://github.com/Kinto/kinto-admin/pull/77/commits/c180b2bab9fbc7215cd3d954353274769e001809

Natim: Ok :+1:

n1k0:
I'm landing this.

PR #80, opened 2016-05-24
Fixes #73: Added support for Firefox Account authentication.

### LINKED ISSUE #73: Add support for Firefox Account authentication

n1k0:
This is WiP, refs #73.

Note: I needed to upgrade a bunch of dependencies because of a bug in react-router which has been fixed in latest version.

I've deployed the patch [to gh-pages](https://kinto.github.io/kinto-admin/), feedback welcome.


Natim:
Does this request makes it possible to give a link with the basicauth setup for the kinto-admin directly?

n1k0:
> Does this request makes it possible to give a link with the basicauth setup for the kinto-admin directly?

Not atm but it should be possible; but what would be the use case? Also I'm realizing having a public link logging you in automatically might not be a good idea...

n1k0:
Well, okay sorry, I misread your comment (thought you were expecting to share an FxA auth link). So no, a link to setup basic auth automatically is not possible atm; Also that would mean having the password in the URL, which is concerning.

Natim:
Yes but the security model with BasicAuth is to use a VPN to authenticate. So that would let people have a bookmark with their credentials that they can use once connected to the VPN.

n1k0:
> Yes but the security model with BasicAuth is to use a VPN to authenticate. 

This is the security model for one project at Mozilla; should we expand this model to the community?

Natim:
Yes no you are probably right.
Maybe an alternative that can prefill the server url and the username but still ask for the password?

n1k0:
If we can live with having the form prefilled with server and username, excluding the password from the URL, yes that should be possible though. That's a good idea, please file an issue to have this addressed outside of this PR.

Natim:
Yes ok

leplatrem:
Auth worked like a charm \o/

Nit: combo of auth method is not affordant (no visual feedback that it is a combox, and not a field). Idea: Use radios ?
capture d ecran de 2016-05-24 11-42-33

Nit2: not username shown on top menu
capture d ecran de 2016-05-24 11-44-39

Natim:
> Use radios ?

:+1: 

> no username shown on top menu

I have created https://github.com/Kinto/kinto/issues/641 about this one.

n1k0:
> Idea: Use radios ?

Yeah, using radios is much better:



> no username shown on top menu

Good catch!

almet:
neat :)

n1k0:
Addressed @leplatrem nits, updated hosted instance. 

May I have a r+ (or more nits) from someone?

Natim:
You can have a r+ from me yes :)

Natim:
Maybe we can use the fxa:FxID as the username before we display the email?

n1k0:
> Maybe we can use the fxa:FxID as the username before we display the email?

I prefer doing everything at once. Merging this.

vanillajonathan:
Have this functionality been removed?

Natim:
It has not, however you need to activate the kinto-fxa plugin on your service to see it

vanillajonathan:
Ah, I see that it does not seem to be activated on https://kinto.github.io/kinto-admin/
Any reason why it is not activated?

Natim:
You can use OpenID Conenct (Auth0) to try it.

PR #87, opened 2016-05-25
Fixes #85: Collection deletion confirmation prompt.

### LINKED ISSUE #85: Implement deletion confirmation form to collections
We should implement the same UI for deleting collections as the one we landed for buckets in #83.

n1k0:
Refs #85.

No component tests waiting for functional tests as discussed with @leplatrem.

Btw, r=? @leplatrem :D

### DIFF
--- a/scripts/components/CollectionEdit.js
+++ b/scripts/components/CollectionEdit.js
@@ -1,8 +1,31 @@
 import React, { Component } from "react";
+import Form from "react-jsonschema-form";
 
 import CollectionForm from "./CollectionForm";
 import Spinner from "./Spinner";
 
+
+const deleteSchema = {
+  type: "string",
+  title: "Please enter the collection name to delete as a confirmation",
+};
+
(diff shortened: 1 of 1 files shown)


scripts/components/CollectionEdit.js
```diff
@@ -1,26 +1,50 @@
...
+  const validate = (formData, errors) => {
+    if (formData !== cid) {
+      errors.addError("The collection name does not match.");
+    }
+    return errors;
+  };
+  return (
+    <Form
+      schema={deleteSchema}
+      validate={validate}
```
Natim: So you actually added the validation :+1:

Natim:
LGTM

scripts/components/CollectionEdit.js
n1k0: [Yes!](https://github.com/mozilla-services/react-jsonschema-form/pull/163)

Though there's no way to define custom validation for userland collections in the admin just yet. Here it's only used internally.

PR #91, opened 2016-05-30
Handle nested keys for the displayFields.

Natim:
- [x] Nested behavior
- [x] Add tests

Early feedback from @n1k0 (can you point me where I should add the tests? Is it possible to unittest recordField?)

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,45 @@ browser needs.
 
 Instead, run the development server, as above.
 
+### How to display a nested field value using the collection displayFields property?
+
+Use the pointed notation.
+
+For example, if you have the following record:
+
+```
+{
+  "data": {
+    "attachement": {
+      "filename": "font.ttf"
+    }
+  }
+}
+```
+
+You can use ``attachment.filename``.
+
+We tried our best to make it work with properties having dots in their name.
+
+For instance:
+
+```
+{
+  "data": {
+    "target": {
+      "proof.hash": "abcd",
+    "merkle.tree": {
+         "file.name": "foobar"
+      }
+    }
+  }
+}
+
+```
+
+If you use ``target.merkle.tree.file.name`` it will render the string
+``foobar`` and ``target.proof.hash`` will render ``abcd``.
+
 ## License
 
 Apache Licence Version 2.0
--- a/README.md
+++ b/README.md
@@ -96,7 +96,7 @@ Instead, run the development server, as above.
 
 ### How to display a nested field value using the collection displayFields property?
 
-Use the pointed notation.
+Use the *dot* notation.
 
 For example, if you have the following record:
 
@@ -110,7 +110,7 @@ For example, if you have the following record:
 }
 ```
 
-You can use ``attachment.filename``.
+You can use `attachment.filename`.
 
 We tried our best to make it work with properties having dots in their name.
 
@@ -121,7 +121,7 @@ For instance:
   "data": {
     "target": {
       "proof.hash": "abcd",
-    "merkle.tree": {
+      "merkle.tree": {
          "file.name": "foobar"
       }
     }
@@ -130,8 +130,8 @@ For instance:
 
 ```
 
-If you use ``target.merkle.tree.file.name`` it will render the string
-``foobar`` and ``target.proof.hash`` will render ``abcd``.
+If you use `target.merkle.tree.file.name` it will render the string
+`foobar` and `target.proof.hash` will render `abcd`.
 
 ## License
 
--- a/scripts/components/CollectionList.js
+++ b/scripts/components/CollectionList.js
@@ -1,7 +1,7 @@
 import React, { Component } from "react";
 import { Link } from "react-router";
 
-import { recordField } from "../utils";
+import { renderDisplayField } from "../utils";
 import Spinner from "./Spinner";
 
 
@@ -39,7 +39,7 @@ class Row extends Component {
     return <tr onDoubleClick={this.onDoubleClick.bind(this)}>
       {
         displayFields.map((displayField, index) => {
-          return <td key={index}>{recordField(displayField, this.props.record)}</td>;
+          return <td key={index}>{renderDisplayField(record, displayField)}</td>;
         })
       }
       <td className="lastmod">{this.lastModified}</td>
--- a/scripts/components/CollectionList.js
+++ b/scripts/components/CollectionList.js
@@ -1,7 +1,7 @@
 import React, { Component } from "react";
 import { Link } from "react-router";
 
-import { cleanRecord } from "../utils";
+import { cleanRecord, recordField } from "../utils";
 import Spinner from "./Spinner";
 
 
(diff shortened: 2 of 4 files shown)


n1k0:
I'd probably move the whole *recordField* logic to the *utils* module and add unit tests for it.

Natim:
Ok let me do that then :+1:

Natim:
@n1k0 now ready for review.

Natim:
I added a commit to handle more complex key matching but I don't know if we want to support it.

See https://github.com/Kinto/kinto-admin/pull/91/commits/0f8ba91938f1bab0924f78ae4daeb942e831f2ff

scripts/components/CollectionList.js
```diff
@@ -34,29 +34,12 @@ class Row extends Component {
...
-    return "<unknown>";
-  }
-
   render() {
     const {bid, cid, record, displayFields} = this.props;
     return <tr onDoubleClick={this.onDoubleClick.bind(this)}>
       {
         displayFields.map((displayField, index) => {
-          return <td key={index}>{this.recordField(displayField)}</td>;
+          return <td key={index}>{recordField(displayField, this.props.record)}</td>;
```
n1k0: Nit: `record` is already destructured and available, no need to get it from `this.props`.

scripts/utils.js
```diff
@@ -28,3 +30,48 @@ export function validJSON(string) {
 export function cleanRecord(record) {
   return omit(record, ["id", "schema", "last_modified"]);
 }
+
+export function recordField(displayField, record) {
```
n1k0: How about renaming this and updating the signature to something more meaningful like `renderDisplayField(record, displayField)`?

scripts/utils.js
```diff
@@ -28,3 +30,48 @@ export function validJSON(string) {
...
+    } else {
+      return String(field);
+    }
+  } else if (displayField === "__json") {
+    return <code>{JSON.stringify(cleanRecord(record))}</code>;
+  } else if (displayField.indexOf(".") !== -1) {
+    const fields = displayField.split(".");
+
+    if (record.hasOwnProperty(fields[0])) {
+      return recordField(fields.splice(1).join("."), record[fields[0]]);
```
n1k0: As we often return early we can probably get rid of a bunch of nested if/else statements.

test/utils_test.js
```diff
@@ -0,0 +1,59 @@
+import { expect } from "chai";
+
+import {cleanRecord, recordField} from "../scripts/utils";
+
+describe("cleanRecord", () => {
```
n1k0: Thanks :)

scripts/utils.js
n1k0: Also be careful as `splice` mutates field; it's no big deal here but I'd rather see `slice` used instead.

test/utils_test.js
```diff
@@ -0,0 +1,59 @@
...
+      "supported.strange.nested": {"tree": "foobar"}
+    };
+  });
+  
+  it("should return the title field as a string", () => {
+    expect(recordField("title", record)).to.eql("I am a title");
+  });
+
+  it("should return the extras field as a JSON string", () => {
+    expect(recordField("extras", record)).to.eql('{"foo":"bar","foobar":"foo"}');
```
n1k0: Nit: `.eql(JSON.stringify(record.extras))`

scripts/utils.js
```diff
@@ -28,3 +30,48 @@ export function validJSON(string) {
...
+    } else {
+      let biggestCandidate = [];
+      let nextCandidate = [];
+      let candidates = Object.keys(record).filter((key) => {
+        return key.indexOf(fields[0]) === 0;
+      });
+
+      for (let key in candidates) {
```
n1k0: It's very strange to iterate over an array using for/in

scripts/utils.js
```diff
@@ -28,3 +30,48 @@ export function validJSON(string) {
...
+          }
+        }
+        if (nextCandidate.length > biggestCandidate.length) {
+          biggestCandidate = nextCandidate;
+        }
+      }
+
+      const key = biggestCandidate.join(".");
+      return recordField(fields.splice(biggestCandidate.length).join("."),
+                         record[key]);
```
n1k0: I'm having a tough time figuring out the actual intent of this code, so we might want to add a few comments there.

scripts/utils.js
Natim: Yes I should use "for/of" here

scripts/utils.js
n1k0: Also I'm realizing this is a useful & powerful feature which should be documented *somewhere*; it's probably worth adding a section to the docs about this.

n1k0: Or ideally a `ui:help` uiSchema directive for the `displayField` widget, so we get contextual help for free.

scripts/utils.js
```diff
@@ -28,3 +30,48 @@ export function validJSON(string) {
...
+        for (let part of fields) {
+          var candidate = nextCandidate.concat([part]).join(".");
+          if (candidates[key].indexOf(nextCandidate.concat([part]).join(".")) !== -1) {
+            nextCandidate.push(part);
+          }
+        }
+        if (nextCandidate.length > biggestCandidate.length) {
+          biggestCandidate = nextCandidate;
+        }
+      }
```
n1k0: I suspect thje code would gain in clarity if this logic is moved to a dedicated function.

README.md
```diff
@@ -94,6 +94,45 @@ browser needs.
...
+We tried our best to make it work with properties having dots in their name.
+
+For instance:
+
+```
+{
+  "data": {
+    "target": {
+      "proof.hash": "abcd",
+    "merkle.tree": {
```
n1k0: Broken indentation and json structure.

README.md
```diff
@@ -94,6 +94,45 @@ browser needs.
 
 Instead, run the development server, as above.
 
+### How to display a nested field value using the collection displayFields property?
+
+Use the pointed notation.
```
n1k0: Nit: the *dot* notation.

README.md
```diff
@@ -94,6 +94,45 @@ browser needs.
...
+         "file.name": "foobar"
+      }
+    }
+  }
+}
+
+```
+
+If you use ``target.merkle.tree.file.name`` it will render the string
+``foobar`` and ``target.proof.hash`` will render ``abcd``.
```
n1k0: Markdown inline code only requires a single backtick.

test/utils_test.js
```diff
@@ -0,0 +1,68 @@
...
+    };
+  });
+  
+  it("should return the title field as a string", () => {
+    expect(renderDisplayField(record, "title")).to.eql("I am a title");
+  });
+
+  it("should return the extras field as a JSON string", () => {
+    expect(renderDisplayField(record, "extras")).to.eql(
+      JSON.stringify({"foo": "bar", "foobar": "foo"}));
```
n1k0: Nit: here and elsewhere in other tests it would be good to always keep this formalism for consistency and to avoid having lines over 80c:

```js
expect(thing)
  .eql(expectation);
```

n1k0:
Ok this looks good to me, great feature. r+ with the remaining nit addressed.

Natim:
Thank you for your great review :)

Issue #93, opened 2016-05-30
Add ui:help for the displayField field in the collection properties form


n1k0:
Spawned from #91.

Natim:
Is a placeholder sufficient? (I know I am lazy)

![placeholder](https://cloud.githubusercontent.com/assets/229453/15654064/58d32ce4-2691-11e6-9747-61fda325b1fe.png)

n1k0:
Yes!

Natim:
I did the two solutions in https://github.com/Kinto/kinto-admin/pull/94
Tell me which one you prefer and I will remove the unwanted commit.

n1k0:
Closed by #94.

PR #94, opened 2016-05-30
Add a field description.

Natim:
First attempt on fixing #93 

It is not exactly a ``ui:help`` yet: 

placeholder


Natim:
With the ui:help it looks like that:

ui help

Any help on the wording?

n1k0:
I really prefer the placeholder approach. Simple yet super efficient.

Natim:
Ok here you go :)

PR #90, opened 2016-05-26
Bucket data edition.

n1k0:
Refs #84. This basically allows editing bucket data (JSON properties) from the UI.

The patch is larger than expected because:

- it introduces a new store, actions and sagas for storing and handling the current selected bucket state;
- it refactors the way we preload resources when the current route changes;
- saga tests are really verbose.

#### Todo from @leplatrem 

- [x] Update bucket and collection name json schema field to add an appropriate validation regexp
- [x] Avoid stacking up error notifications

PR #97, opened 2016-05-30
Fixes #95: Added missing bucket prop to dispatched action payload.

### LINKED ISSUE #95: Loosing collection records when navigating back to collection list
#### Steps to reproduce

1. log in
2. navigate to some collection properties form
3. navigate to its list of records

#### Expected

The list of records should be there

#### Actual

The list of records is empty even if records exist for this collection.

n1k0:
This solves two issues now #90 has landed:

- We were loosing the list of records when navigating from collection edition to the list; 
- We were missing the bucket name in the title when editing collection properties.

### DIFF
--- a/scripts/actions/collection.js
+++ b/scripts/actions/collection.js
@@ -20,8 +20,8 @@ export function resetCollection() {
   return {type: COLLECTION_RESET};
 }
 
-export function collectionLoadSuccess(properties) {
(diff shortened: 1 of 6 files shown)


n1k0:
Deployed the patch to gh-pages. r=? @Natim

n1k0:
Landing this with r=me.

Issue #66, opened 2016-05-19
Render the uuid in the records list


n1k0:
Spawned from the comments in #65.

PR #123, opened 2016-06-02
Validate provided displayFields against the json schema.

n1k0:
This adds validation to the Collection creation/edition form to ensure that entered _displayFields_ actually match the provided json schema.

r=? @Natim @leplatrem

### DIFF
--- a/scripts/utils.js
+++ b/scripts/utils.js
@@ -72,6 +72,16 @@ export function validateSchema(jsonSchema) {
   return schema;
 }
 
+export function validateDisplayFields(schema, displayFields) {
+  const {properties} = JSON.parse(schema);
+  const fieldRoots = displayFields.map((field) => field.split(".")[0]);
+  fieldRoots.forEach((root) => {
+    if (!properties.hasOwnProperty(root)) {
+      throw `The JSON schema does not define a ${root} field`;
+    }
+  });
+}
+
 export function cleanRecord(record) {
   return omit(record, ["id", "schema", "last_modified"]);
 }
--- a/scripts/utils.js
+++ b/scripts/utils.js
@@ -72,14 +72,26 @@ export function validateSchema(jsonSchema) {
   return schema;
 }
 
+function extractSchemaPaths(schema, prefix="") {
+  const paths = [];
+  if (!schema.hasOwnProperty("properties")) {
+    return paths;
+  }
+  return Object.keys(schema.properties).reduce((paths, key) => {
+    const path = (prefix ? prefix + "." : prefix) + key;
+    return paths.concat(path, extractSchemaPaths(schema.properties[key], path));
+  }, paths);
+}
+
 export function validateDisplayFields(schema, displayFields) {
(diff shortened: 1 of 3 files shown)


n1k0:
Deployed to gh-pages.

scripts/utils.js
```diff
@@ -72,6 +72,28 @@ export function validateSchema(jsonSchema) {
...
 }
 
+function extractSchemaPaths(schema, prefix="") {
+  const paths = [];
+  if (!schema.hasOwnProperty("properties")) {
+    return paths;
+  }
+  return Object.keys(schema.properties).reduce((paths, key) => {
+    const path = (prefix ? prefix + "." : prefix) + key;
+    return paths.concat(path, extractSchemaPaths(schema.properties[key], path));
```
Natim: Genius

Natim:
The code LGTM, I will have a look at the usability on gh-pages.

Natim:
Ah you need to allow to display id and last_modified fields too.

Natim:
I am wondering how you plan to handle cases where you have a displayField that are not in the schema but can be in the record?

n1k0:
> Ah you need to allow to display id and last_modified fields too.

`last_modified` is displayed by default in a dedicated column in the list. Remains `id` which should be addressed by #66 (we actually always want the id rendered in the list).

> I am wondering how you plan to handle cases where you have a displayField that are not in the schema but can be in the record?

Hmm, that's a strange use case but that may happen indeed. I can't infer these, so we should probably discard this patch as I can't think of any technical solution to address this.

n1k0:
Now I think a bit more about it, we could split the current displayField field into two different ones:

- *schemaFields* would select properties part of the schema;
- *recordFields* would allow defining other properties.

The *schemaFields* field could be a list of checkboxes listing all the available field paths (see the patch), and users would simply cherry-pick the ones they want to render in the table view.

Bunch of work, but may be worth it.

Natim:
Or you can add a warning (this field is not in the schema, are you sure you didn't made a typo?)

n1k0:
Yeah that may work too, and it's much simpler.

Natim:
What is the status of this? Should we close it?

n1k0:
Yes, probably.

n1k0:
#412 is probably a good hint that we should work on this again.

scripts/utils.js
```diff
@@ -72,6 +72,28 @@ export function validateSchema(jsonSchema) {
   return schema;
 }
 
+function extractSchemaPaths(schema, prefix="") {
+  const paths = [];
+  if (!schema.hasOwnProperty("properties")) {
+    return paths;
+  }
+  return Object.keys(schema.properties).reduce((paths, key) => {
+    const path = (prefix ? prefix + "." : prefix) + key;
```
itaisteinherz:
```suggestion
    const path = `${prefix}${prefix ? "." : ""}${key}`;
```

leplatrem:
Pull Request TTL timeout.

Issue #124, opened 2016-06-02
If the user have got a 403 on the bucket, it cannot open the collection.


Natim:
![403-on-bucket](https://cloud.githubusercontent.com/assets/229453/15743955/4318755a-28c9-11e6-9338-e6f16d45c868.png)

![error-undefined](https://cloud.githubusercontent.com/assets/229453/15743957/45819bfa-28c9-11e6-89d1-b7c684285624.png)

PR #125, opened 2016-06-02
Changes to allow to load subresources on 403.

Natim:
First step in trying to address #124

### DIFF
--- a/scripts/reducers/bucket.js
+++ b/scripts/reducers/bucket.js
@@ -18,7 +18,10 @@ export function bucket(state = INITIAL_STATE, action) {
       return {...state, busy: action.busy};
     }
     case BUCKET_LOAD_SUCCESS: {
-      const {data} = action;
+      let {bid, data} = action;
+      if (typeof data === "undefined") {
+        data = {id: bid};
+      }
       return {
         ...state,
         data: omit(data, ["id", "last_modified"]),
--- a/scripts/sagas/bucket.js
+++ b/scripts/sagas/bucket.js
@@ -31,7 +31,8 @@ export function* loadBucket(bid) {
   try {
     const bucket = getBucket(bid);
     yield put(bucketBusy(true));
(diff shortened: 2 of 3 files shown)


scripts/reducers/bucket.js
```diff
@@ -18,7 +18,10 @@ export function bucket(state = INITIAL_STATE, action) {
       return {...state, busy: action.busy};
     }
     case BUCKET_LOAD_SUCCESS: {
-      const {data} = action;
+      let {bid, data} = action;
+      if (typeof data === "undefined") {
```
n1k0: I think this should be processed in the reducer but rather in the `listBuckets` saga, when we receive the 403.

Natim: We should actually improve that patch because since we have a 403 we actually do not have it in listBuckets.

Natim:
Actually I think the best course of actions would be to let user add buckets and before creating it we should try to list its collection.

So that if I want to open an existing bucket I do not have permission on but I have permission on its collection, I can use the create bucket or hidden bucket button to load it nevertheless.

Natim:
For instance I cannot open https://form:c78110f1e7514ff5afd89ee28a568872@kinto.notmyidea.org/v1/buckets/formbuilder (403) but I can list https://form:c78110f1e7514ff5afd89ee28a568872@kinto.notmyidea.org/v1/buckets/formbuilder/collections

n1k0:
Nopte that all of this would be solved if we had https://github.com/Kinto/kinto/issues/600, which may have just turned P1 with this issue ;)

scripts/reducers/bucket.js
n1k0: Yeah I was more thinking of `loadRoute` actually.

n1k0:
Fixed by #128.

Natim:
Ok thanks.

Issue #131, opened 2016-06-06
Prevent rendering forms you're not allwoed to fill


n1k0:
#### Todo

- [x] Check permission when editing a bucket (#134) 
- [x] Check permission when editing a collection 
- [x] Check permission when editing a record 

#### The issue

![](http://i.imgur.com/cehJ6tr.png)

This is super frustrating :)

Natim:
If you have the read permission you should be able to display them.
However the submit button should be missing and the fields readonly.

n1k0:
Yes indeed. That plus a clear warning at the top and we're good to go.

n1k0:
Fixed by 135.

PR #134, opened 2016-06-06
Refs #131: Mark bucket edition form as readonly if user doesn't have write permission

n1k0:
This refs #131, only dealing with bucket forms atm. Next step will be to deal with collection forms.

Deployed to gh-pages. r=? @Natim


Natim:
Do you plan on doing the same thing for collections in another PR?

LGTM

n1k0:
yes

PR #135, opened 2016-06-06
Refs #131: Check permission when editing a collection.

n1k0:
Refs #131, r=? @Natim

Deployed to gh-pages.


Natim:
Great \o/

Natim:
I don't know if it is related but I cannot log in.
I have the following error:

yep

Natim:
r+ In Private mode I don't reproduce so it should be a cache issue on my side.

PR #136, opened 2016-06-07
Experimental use of flow.

n1k0:
Just wanted to experiment with [flow](http://flowtype.org/), but it has already found a well hidden bug!


scripts/utils.js
```diff
@@ -180,12 +171,12 @@ export function extractFileInfo(dataURL) {
...
 
-export function createFormData(record) {
+export function createFormData(record: Object) {
   const {FormData} = window;
   const attachment = record.__attachment__; // data-url
   const {blob, name} = extractFileInfo(attachment);
   const formData = new FormData();
   formData.append("attachment", blob, name);
-  formData.append("data", JSON.stringify(omit(record, "__attachment__")));
+  formData.append("data", JSON.stringify(omit(record, ["__attachment__"])));
```
n1k0: Look at that!

Natim: Ah bah ce sont des choses qui arrivent !

Natim:
r+

scripts/client.js
```diff
@@ -1,47 +1,69 @@
+/* @flow */
+
 import KintoClient from "kinto-client";
 import endpoint from "kinto-client/lib/endpoint";
 import { isHTTPok } from "./utils";
 
 
-let client;
+let client: ?KintoClient;
```
glasserc: OMG nullable types! Just when I had given up hope!

n1k0: Haha yeah, I've already spotted a few bugs thanks to this :)

n1k0:
Ok guys I'm exhausted adding types and the PR is bloating up :D

All in all it fixed a few quirks already and the patch is mostly type declarations in actions, reducers, client and utils modules. It still misses types for sagas and components, but the good news is we can iterate.

I'd love a review from @glasserc who expressed interest for types in the past ;) but I'm taking any feedback really /cc @almet @Natim @leplatrem.

scripts/client.js
```diff
@@ -1,47 +1,72 @@
...
-  const {authType, credentials} = session;
+function getAuthHeader(session: Object): ?string {
+  const {authType, credentials}: {
+    authType: string,
+    credentials: Object
+  } = session;
   switch(authType) {
     case "fxa": {
-      const {token} = credentials;
+      const {token}: {token: string} = credentials;
```
glasserc: I'm not crazy about the fact that the fieldname appears twice. Is it possible to do something more like `const {token: string} = credentials` ?

scripts/client.js
```diff
@@ -1,47 +1,72 @@
...
+      } = credentials;
       return "Basic " + btoa([username, password].join(":"));
     }
   }
 }
 
-export function setupClient(session) {
-  const {server} = session;
+export function setupClient(session: Object): KintoClient {
+  const {server}: {server: string} = session;
```
glasserc: For some of these duplicated types, it seems like it would be better to just move the type into the function declaration, i.e.

```javascript
export function setupClient(session: {server: string}): KintoClient {
  const {server} = session;
```

glasserc: Hmm, on second thought, this might not be possible because of the call to `getAuthHeader` below, which has two very different signatures depending on context..

scripts/types.js
```diff
@@ -0,0 +1,101 @@
+/* @flow */
+
+export type Action = {
+  type: string
+};
```
glasserc: Further work here could involve https://flowtype.org/docs/disjoint-unions.html but I don't think it's necessary for this PR

test/sagas/session_test.js
```diff
@@ -69,7 +69,7 @@ describe("session sagas", () => {
     });
 
     it("should reset the client", () => {
-      expect(getClient()).to.be.a("null");
+      expect(() => getClient()).to.Throw(Error, /not configured/);
```
glasserc: Is `Throw` supposed to be capitalized here?

glasserc:
In general, I like this! I'm very impressed with `flow` based on what I see here. It's awesome, if a little confusing, to see JavaScript with a type system stronger than go's :)

scripts/client.js
n1k0: Sadly that's the only type declaration syntax supported by flow for typed object destructured assignments.

scripts/client.js
n1k0: Indeed, though I realize I could add types for these.

scripts/types.js
n1k0: My thoughts exactly, will do in a followup PR.

n1k0:
Ok I'm gonna land this.

PR #137, opened 2016-06-09
Prepare v1.0.0.

n1k0:
### Checklist

- [x] Update all dependencies to their latest stable versions
- [x] Check docs
- [x] Investigate how we could deploy tagged stable versions in gh-pages subdirectories, eg `kinto.github.io/kinto-admin/1.0.0/`
- [x] Create release notes

### Release notes

This is the first major stable version of kinto-admin. Since latest v0.16.1, the codebase has been entirely rewritten to switch from an offline-first approach to a simpler live HTTP one. We really wish we could have introduced such a breaking change more gradually, but shifting that heavily can hardly be progressive to be honest. 

Now, the admin is more powerful and feature-complete than ever; here are a few highlights: 

- You don't need to provide a configuration file anymore as resources can be discovered and browsed directly from the UI;
- You can now create buckets and collections from the admin, configure their metadata (eg. collection schemas) and even delete them;
- We've added support for [Firefox Account](https://accounts.firefox.com/) authentication along the *BasicAuth* one; we're planning on introducing more alternative OAuth2 auth providers in the future;
- We've added native support for [kinto-attachments](https://kinto.github.io/kinto-attachment/), which means you can now attach files to records and post them to the server if the capability is available;

We really hope you'll enjoy this new version as much as we've loved building it. Of course there are and will be bugs, but you already know [where to file them](https://github.com/Kinto/kinto-admin/issues) :)


Natim:
This is it!

n1k0:
Haha yeah, well there are many issues remaining but I *think* we can live with them as the admin is globally useable and its general architecture is unlikely to change anytime soon.

n1k0:
Found a trick to deploy a stable tag to gh-pages in https://github.com/Kinto/kinto-admin/pull/137/commits/c0e58853f4d29a2b0593ac5ecebb3e5d966d09d1, updated https://github.com/Kinto/kinto-admin/wiki/Releasing accordingly

n1k0:
I'm gonna ask @glasserc a quick review so I can eventually release tomorrow :)

package.json
```diff
@@ -25,61 +29,62 @@
     "node": ">=4"
   },
   "dependencies": {
-    "btoa": "^1.1.2",
-    "codemirror": "^5.13.2",
     "express": "^4.13.3",
-    "extract-text-webpack-plugin": "^1.0.1",
-    "filesize": "^3.3.0",
-    "kinto-client": "^0.8.3",
```
glasserc: Just to confirm, we're saying that the only true "dependencies" of this project are the ones we use to serve it, and everything that is part of the code that we actually use when we run it goes into "devDependencies"? Seems strange, but OK.

glasserc:
Otherwise, looks OK to me. r+wc

package.json
n1k0: Yeah, would be odd to package dependencies not actually used by the package. The npm package only exposes an executable capable of serving static assets, so we should declare the dependencies for doing so, and nothing more.

In the future if we expose an API to extend the app capabilities, that will be another story, and we'll reintroduce the actual dependencies for it. Right now, it just wouldn't make any sense.

PR #143, opened 2016-06-10
Fixes #138: Validate uiSchema.

### LINKED ISSUE #138: [minor] Validate uiSchema during collection creation/edition
Error appears while adding record otherwise:
capture d ecran de 2016-06-10 10-23-13

It is enough for our current use-cases, but could be helpful for new comers

n1k0:
r=? @leplatrem

### DIFF
--- a/scripts/utils.js
+++ b/scripts/utils.js
@@ -95,7 +95,7 @@ export function validateUiSchema(jsonUiSchema: string, jsonSchema: string) {
       },
       {
         test: () => arrayId(order) === arrayId(properties),
-        error: "The ui:order list should match schema properties",
+        error: "The ui:order directive should list all schema properties",
       }
     ]);
   }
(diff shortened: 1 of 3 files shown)


n1k0:
Falling back to @glasserc or @Natim for review as @leplatrem is possibly on PTO.

scripts/utils.js
```diff
@@ -70,6 +70,43 @@ export function validateSchema(jsonSchema: string) {
...
+    const properties: string[] = Object.keys(schema.properties);
+    const arrayId = (array: string[]): string => array.slice().sort().toString();
+    checks = checks.concat([
+      {
+        test: () => Array.isArray(order),
+        error: "The uiSchema ui:order directive isn't an array",
+      },
+      {
+        test: () => arrayId(order) === arrayId(properties),
+        error: "The ui:order list should match schema properties",
```
Natim: I would say that the ui:order should list all schema properties

Natim:
r+wc

PR #144, opened 2016-06-10
Fixes #132: Ensure resetting session dependent sagas.

### LINKED ISSUE #132: Reload watchers for serverInfo when refetched with new information
Reproduction steps:

1. Log in
2. Create a record
3. Log out
4. Log in with new server information (eg. new remote server)
5. Previous server information are used to check for processing attachments

n1k0:
Refs #132. Basically when we start watching for record create/update events, we need to check that the server actually supports attachments; to achieve that we wait for the "server information loaded" event to be triggered, then we wait forever for record create/update events. The problem is when you log out and log in again using a different Kinto server, you're still stuck within that neverending loop, and you're processing files using the previous server information. Bad.

So the fix consists on resetting the neverending watchers for create/update record events anytime we receive new server information. Which is precisely the purpose of `takeLatest` http://yelouafi.github.io/redux-saga/docs/api/index.html#takelatestpattern-saga-args

### DIFF
--- a/scripts/sagas/collection.js
+++ b/scripts/sagas/collection.js
@@ -198,8 +198,9 @@ export function* watchCollectionRecords() {
   }
 }
 
-export function* watchRecordCreate() {
-  const {serverInfo} = yield take(SESSION_SERVERINFO_SUCCESS);
+export function* watchRecordCreate(serverInfoAction) {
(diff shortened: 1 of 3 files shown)


glasserc:
LGTM, although I have to say I don't 100% understand sagas or how they're used in this project.

PR #146, opened 2016-06-12
Fixes #92: Added simple browser test suite

### LINKED ISSUE #92: Add a minimal browser test suite
As the admin keeps growing and adding features, we need to start thinking about adding end to end browser tests to ensure some critical flows don't break.

n1k0:
Refs #92. These are functional, end to end browser tests for a few basic features of the Kinto admin.

[](https://www.youtube.com/watch?v=LbuGz2PdlUk)

(click the image to see it in action)

### Todo

- [x] Setup nightmare
- [x] Setup static web server
- [x] Setup kinto test server
- [x] Add a few browser testing scenarios
- [x] Add db checks for tests involving server transactions
- [x] Update docs


n1k0:
I know most of you guys are travelling but feedback and/or reviews are welcome :) @glasserc @leplatrem @Natim @almet

n1k0:
It looks like the few intermittent failures we get are because of xvfb messing with electron/chromium https://github.com/segmentio/nightmare/issues/224. Now the question si, can we live with these until this is fixed?

n1k0:
Another option would be to exclude the browser test suite from the Travis build matrix, and leave the responsibility of running it periodically to the developers. I don't like it, but I don't have a plan B. Thoughts?

glasserc:
I've looked at this. It's not what I was expecting. I was expecting some kind of end-to-end test that verifies that clicking the "delete" button actually deletes a record. This is more like a set of unit tests for the browser UI to verify that what the user sees is sane, without regard to whether it actually did any of the things it's showing the user. I don't have a lot of experience with React-based applications so it's hard for me to know whether this is "enough". I'm a little nervous about it, but I guess it's uncommon for the UI to act correctly without actually sending requests and doing things, so this is probably OK.

The code looks sane to me. It's all in JS, so code reuse is easy to understand, which is a lot better than things were the last time I tried to use Selenium.

I guess you have to use `.evaluate` for anything you want to happen "in the browser"? Is this going to be annoying to share? For example, there are two places where you override `confirm()`.

n1k0:
Good point, I'll add checks for ensuring that the database state matches what's rendered in the UI.

glasserc:
I'm also OK with the tests not being mandatory for Travis.

n1k0:
We may want to keep these tests in Travis but allow them to fail https://docs.travis-ci.com/user/customizing-the-build/#Rows-that-are-Allowed-to-Fail

PR #147, opened 2016-06-13
Fixes #82: Allow ordering records list by column.

### LINKED ISSUE #82: (minor) Validate collection id/name
capture d ecran de 2016-05-24 11-50-16

Idea: use the current field to populate a `label` attribute, and slugify to assign the `id` attribute to be used.

n1k0:
Refs #82, this is WiP and ~~doesn't work just yet~~ does work now (I think).

Note: this also fixes an unreported bug where we're loading the list of records **twice** each time we're accessing the related route.

#### Todo

- [x] Basic sorting implementation
- [x] Tests
- [x] Hide sort links for non-schema or nested fields as they're not supported
- [x] Store default sort value in collection data?


scripts/components/CollectionList.js
```diff
@@ -58,6 +58,32 @@ class Row extends Component {
   }
 }
 
+function SortLink(props) {
+  const {dir, column, updateSort} = props;
+  return (
+    <a href="#" className="sort-link" onClick={(event) => {
+      event.preventDefault();
+      updateSort(dir === "up" ? `-${column}` : column);
```
Natim: I don't understand how `up` means `-$column`

n1k0: This is a toggler so we need the action opposite to current state.

n1k0:
I'm now taking early feedback. Deployed the patch to gh-pages. Yeah, we badly need a web designer :cry: 

/cc @glasserc @leplatrem @Natim

Natim:
Let me have a look

Natim:
I think went we activate the ordering it should stay ASC as the arrow shows it.

leplatrem:
Nice! GG!

In term of UX, I have one remark: it is a bit troubling that the whole array disappears behind the loader.

If feasible and not too costly, keeping the headers row would help keeping the mental context of the click :)

n1k0:
> In term of UX, I have one remark: it is a bit troubling that the whole array disappears behind the loader.

I agree though that will be addressed in a followup PR as there's a bunch spinner mess to clean and this would probably bloat up this PR :)

n1k0:
Patch is complete I think. May I now ask for a review of the code? :) /cc @leplatrem @Natim @glasserc

n1k0:
I've been self-reviewing this patch a lot and am pretty sure it's ready for landing, so I'm gonna land this so I can move on.

PR #149, opened 2016-06-16
Fixes #139: Redirect to coll list after coll update,

### LINKED ISSUE #139: [minor] Redirect to records list after collection update?
When a collection is created, we're redirected to the records list.

It is not the case on collection update.

Is it expected?

n1k0:
r=? @leplatrem


leplatrem:
r+

PR #151, opened 2016-06-16
Upgrade kinto-client to kinto-http 1.0.0.

n1k0:
r=? @Natim


Natim:
You need to change all calls to kinto-client into kinto-http too.

n1k0:
done

Natim:
r+

PR #154, opened 2016-06-20
Fixes #153: Improve record list loading UX.

### LINKED ISSUE #153: Avoid blinking spinner when sorting the records list
Followup from #147.

n1k0:
I went with greying the rows while fetching the reordered list from the server. Deployed to gh-pages. r=? @leplatrem

Note: this also removes a bunch of unnecessary crap from sagas.

### DIFF
--- a/scripts/sagas/collection.js
+++ b/scripts/sagas/collection.js
@@ -1,5 +1,4 @@
 import { push as updatePath } from "react-router-redux";
-import { takeLatest } from "redux-saga";
 import { call, take, fork, put } from "redux-saga/effects";
 import { v4 as uuid } from "uuid";
 import { createFormData } from "../utils";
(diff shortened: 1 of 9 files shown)


scripts/sagas/collection.js
```diff
@@ -39,13 +37,10 @@ export function* deleteAttachment(bid, cid, rid) {
 export function* listRecords(bid, cid, sort="-last_modified") {
   try {
     const coll = getCollection(bid, cid);
-    yield put(actions.collectionBusy(true));
```
n1k0: Note: I realized that dealing with these is much more easy to achieve in the reducers directly. That's how you learn I guess. Will need to do the same with other sagas.

scripts/sagas/collection.js
```diff
@@ -184,20 +179,12 @@ function shouldProcessAttachment(serverInfo, records) {
...
-  // collection data.
   while(true) { // eslint-disable-line
     const {bid, cid, sort} = yield take(COLLECTION_RECORDS_REQUEST);
     yield fork(listRecords, bid, cid, sort);
   }
 }
 
-export function* watchResetListRecords() {
-  // On each route update, reset the records loader
-  yield* takeLatest(ROUTE_LOAD_SUCCESS, watchListRecords);
```
n1k0: Note: I realized that while this works, we're not ensuring the route data are loaded prior to loading the list of records anymore, which may (or may not) introduce race conditions. Need to think about it a little more.

leplatrem:
pair reviewed! r+

PR #156, opened 2016-06-22
Saga refactor

n1k0:
WiP. This patch is about simplifying sagas and making them more powerful and easy to use, by passing them a simple way to access current store state. This will be super useful when we're gonna need to contextually check current state (eg. config, user permissions, etc) during async flows.

The first saga fully leveraging this feature is `updateRecord`, which contextually checks for attachment support. That makes a "rather" short patch to grab feedback from. If the approach is good, I can migrate other sagas as well.

If you look carefully at the `updateRecord` signature, you'll see it accepts `getState` and `action` arguments; this is likely to become the generic signature for *all* sagas, which makes adding new ones a no brainer.

Feedback=? @leplatrem as we discussed this already.

### DIFF
--- a/scripts/sagas/index.js
+++ b/scripts/sagas/index.js
@@ -8,7 +8,7 @@ import * as bucketSagas from "./bucket";
 import * as collectionSagas from "./collection";
 
 
-export default function* rootSaga() {
+export default function* rootSaga(getState) {
   yield [
     // session
     fork(sessionSagas.watchSessionSetup),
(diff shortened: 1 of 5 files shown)


scripts/sagas/index.js
```diff
@@ -8,7 +8,7 @@ import * as bucketSagas from "./bucket";
 import * as collectionSagas from "./collection";
 
 
-export default function* rootSaga() {
+export default function* rootSaga(getState) {
```
n1k0: Add a comment to explain where does `getState` come from and why.

scripts/store/configureStore.js
```diff
@@ -14,6 +14,7 @@ const finalCreateStore = compose(
 
 export default function configureStore(initialState) {
   const store = finalCreateStore(rootReducer, initialState);
-  sagaMiddleware.run(rootSaga);
+  const getState = store.getState.bind(store);
```
n1k0: Add a comment explaining that we want all saga to have access to the store state so we bind it to the rootSaga.

n1k0:
Pair reviewed with @leplatrem

glasserc:
I know I'm late to the party, but why `getState` instead of just `state` itself?

glasserc:
Otherwise :+1:, this is much cleaner.

n1k0:
> I know I'm late to the party, but why getState instead of just state itself?

Because the `getState` arg is a function, we found this name was conveying that more efficiently. The immediate advantage of it being a function is that you'll always retrieve the freshest data when calling it comparatively to passing a "frozen" state object, which may be already outdated at the time you use it.

glasserc:
OK, thanks. I thought it might be something like that, but I thought the state never changes. But now I'm rereading the code and I see that I misread it the first time -- the `store` never changes, but the state can change any time.

PR #157, opened 2016-06-23
Remove watcher level

n1k0:
Paired with @leplatrem


n1k0:
Just thinking about it, just let me deploy to gh-pages so we ensure nothing is functionaly broken (shouldn't be but, well, you know).

PR #158, opened 2016-06-24
Remove session watchers

n1k0:
This is pursuing the work started with #156 and #157, moves the session saga watchers to the root saga and update the saga signatures to receive `getState` and the full action payload.

r=? @glasserc @Natim

### DIFF
--- a/test/sagas/index_test.js
+++ b/test/sagas/index_test.js
@@ -1,86 +1,53 @@
 import { expect } from "chai";
 import sinon from "sinon";
-import { fork } from "redux-saga/effects";
 
 import { SESSION_SERVERINFO_SUCCESS } from "../../scripts/constants";
-const routeSagas = require("../../scripts/sagas/route");
+// const routeSagas = require("../../scripts/sagas/route");
 const sessionSagas = require("../../scripts/sagas/session");
 const bucketSagas = require("../../scripts/sagas/bucket");
 const collectionSagas = require("../../scripts/sagas/collection");
-import rootSaga from "../../scripts/sagas";
 import configureStore from "../../scripts/store/configureStore";
 
+import * as sessionActions from "../../scripts/actions/session";
 import * as bucketActions from "../../scripts/actions/bucket";
 
 
+function expectSagaCalled(saga, action) {
+  // Note: the rootSaga function is called by configureStore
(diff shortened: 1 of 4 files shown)


test/sagas/index_test.js
```diff
@@ -1,86 +1,60 @@
...
 
+import * as sessionActions from "../../scripts/actions/session";
 import * as bucketActions from "../../scripts/actions/bucket";
 
 
+function expectSagaCalled(saga, action) {
+  // Note: the rootSaga function is called by configureStore
+  configureStore().dispatch(action);
+
+  expect(saga.firstCall.args[0].name).eql("bound getState");
```
glasserc: This might be a n00b question, but where does this come from? I grepped the repo and googled and didn't find anything.

glasserc:
The code itself looks obviously correct to me, although I don't 100% understand the tests. r+wq

test/sagas/index_test.js
n1k0: Ah, right, I should have put a comment.  

- `saga.firstCall` is a reference to an object describing the first sub call made, and
- `saga.firstCall.args` are the list of args passed to it
- hence `saga.firstCall.args[0]` being the first one, it's our friend `getState`, which is a [bound](https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Objets_globaux/Function/bind) function
- and so `saga.firstCall.args[1]` is the action object

Why can't we test that `saga.firstCall.args[0]` equals `store.getState`? Because `store.getState` is a reference to the prototypal function of the store object while `saga.firstCall.args[0]` is the equivalent of `store.getState.bind(store)` (see `configureStore`), and these two functions cannot really be conpared, the same way these can't:

```js
> function foo(){}
> foo === foo
true
> foo.bind(null) === foo
false
```

So we're using the `name` function property, which in V8 gives you the name prefixed with the word `bound` when it is (the vomitorium is behind the bathroom). That's the only way we've found to check that the first passed argument was our `getState` bound function here. If you have a better plan, I'm all ears!

glasserc: No, that's fine, "it comes from V8" is sufficient for my purposes. I thought you were hand-making a mock or something of the store, but couldn't figure out how.

PR #159, opened 2016-06-27
Remove route saga watchers.

n1k0:
This patch is a next step in the refactoring process started with #156, #157 and #158.


Natim:
:+1: r+

PR #160, opened 2016-06-27
Remove collection watchers

n1k0:
This patch is the final step in the refactoring process started with #156, #157, #158 and #159.

### DIFF
--- a/test/sagas/collection_test.js
+++ b/test/sagas/collection_test.js
@@ -5,13 +5,11 @@ import { v4 as uuid } from "uuid";
 
 import {
   SESSION_SERVERINFO_SUCCESS,
-  RECORD_CREATE_REQUEST,
   RECORD_UPDATE_REQUEST,
   RECORD_DELETE_REQUEST,
   RECORD_BULK_CREATE_REQUEST,
 } from "../../scripts/constants";
 import { notifyError, notifySuccess } from "../../scripts/actions/notifications";
-import * as sessionActions from "../../scripts/actions/session";
 import * as collectionActions from "../../scripts/actions/collection";
 import * as recordActions from "../../scripts/actions/record";
 import * as saga from "../../scripts/sagas/collection";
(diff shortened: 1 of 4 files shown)


n1k0:
r=? @leplatrem @Natim @magopian @glasserc

test/sagas/collection_test.js
```diff
@@ -101,14 +99,27 @@ describe("collection sagas", () => {
...
   describe("createRecord()", () => {
-    describe("Success", () => {
-      let collection, createRecord;
+    let collection;
+
+    before(() => {
+      collection = {createRecord() {}};
+      const bucket = {collection() {return collection;}};
+      setClient({bucket() {return bucket;}});
+    });
```
glasserc: Why is this a `before` and not a `beforeEach`? When you're testing generators, I totally understand why you only want to initialize things once, but I thought it was somehow better practice to use `beforeEach` if you didn't need to initialize things once.

glasserc:
I'm a little nervous about how tightly the tests are tied to the implementation that satisfies them. It feels a little bit like we're just writing the same code twice, once as an implementation and once as a test. But I do see the value, especially in the case of things like wanting to simulate failure.

Anyhow, r+wq :)

test/sagas/collection_test.js
n1k0: It's just as we don't stub anything here, we don't have to reset/revert anything too, so `before` is just enough.

n1k0:
> I'm a little nervous about how tightly the tests are tied to the implementation that satisfies them.

Is this a general remark with the saga tests or with this patch in particular?

I think it's possible to have some generic test helper checking that a given yielded value is part of a sequence, so we could have much less verbose tests, but we'd loose sequential checks, which I think is rather important when testing sequential async flows.

How would you test this better?

glasserc:
It's a general remark with the saga tests. In my opinion, a good test is one that tests behavior against known properties of a correct solution. Instead, these tests just test that the implementation behaves exactly as the implementation does. It feels like there's minimal value in that.

I don't really have any better ideas. At this point, I'd be trying to write integration tests at a coarser level -- for instance, sending a `RECORD_DELETE_REQUEST` and verifying that a record was deleted. Or we could drop this kind of testing and rely on the browser tests you were working on. That would at least examine the behavior of the implementation. But that's outside the scope of this PR.

Another thing that's outside the scope of this PR -- why are you yielding `call`s to `uuid` ? It's not asynchronous, is it? I thought it might be because you wanted to force it to return different results, but then why not just stub/mock the `uuid` function?

n1k0:
> for instance, sending a RECORD_DELETE_REQUEST and verifying that a record was deleted

You mean involving an actual Kinto server running? That would even be functional testing and would get us far beyond unit testing. These unit tests are verbose but they ensure the i/os with the generator state itself are behaving as expected. I don't see anything wrong with this, it's useful to me at least; but I agree I'm finding the tests very verbose. I'll keep thinking about this, but I'm gonna land this patch now.

> why are you yielding calls to uuid ?

Because I'm trying to use `call` for every calls, for consistency and future-proofiness. Also it's easier to test as you don't need to stub it as you mentioned.

PR #161, opened 2016-06-28
List resources a user has access to in the sidebar.

n1k0:
Paired with @magopian, refs https://github.com/Kinto/kinto/pull/653

I didn't deploy to gh-pages because it'd need the feature to be installed on the dev server.

/cc @leplatrem @Natim @glasserc

### DIFF
--- a/scripts/client.js
+++ b/scripts/client.js
@@ -80,3 +80,28 @@ export function requestAttachment(
       throw new Error(`Unable to ${method} attachment: ${message}`);
     });
 }
+
+export function requestPermissions(): Promise {
+  const client: KintoClient = getClient();
+  if (!client) {
+    throw new Error("Client is not configured.");
+  }
+  const {remote, defaultReqOptions}: {
+    remote: string,
+    defaultReqOptions: {
+      headers: Object
+    }
+  } = client;
+  return fetch(`${remote}/permissions`, {
+    headers: defaultReqOptions.headers,
+  })
+    .then((res) => {
+      if (!isHTTPok(res.status)) {
+        throw new Error("HTTP ${status}");
+      }
+      return res.json();
+    })
+    .catch(({message}) => {
+      throw new Error(`Unable to request the permissions endpoint: ${message}`);
+    });
+}
--- a/scripts/sagas/session.js
+++ b/scripts/sagas/session.js
@@ -50,7 +50,7 @@ export function* listBuckets(getState, action) {
       for (const {id} of data) {
         batch.bucket(id).listCollections();
         // XXX future: query the /permissions endpoint to retrieve the other
-        // collections we may have access to bcause they're shared with us
+        // collections we may have access to because they're shared with us
       }
     });
     const buckets = data.map((bucket, index) => {
@@ -68,7 +68,8 @@ export function* listBuckets(getState, action) {
         bucket = {id: permission.bucket_id, collections: []};
         buckets.push(bucket);
       }
-      // Add any missing collection to the current bucket collections list
+      // Add any missing collection to the current bucket collections list; note
+      // that this will expose collections we have shared records within too.
       if ("collection_id" in permission) {
         const collection = bucket.collections.find(x => x.id === permission.collection_id);
         if (!collection) {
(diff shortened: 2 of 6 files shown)


scripts/client.js
```diff
@@ -80,3 +80,28 @@ export function requestAttachment(
...
+
+export function requestPermissions(): Promise {
+  const client: KintoClient = getClient();
+  if (!client) {
+    throw new Error("Client is not configured.");
+  }
+  const {remote, defaultReqOptions}: {
+    remote: string,
+    defaultReqOptions: {
+      headers: Object
```
Natim: Is Object better than {} here?

scripts/client.js
```diff
@@ -80,3 +80,28 @@ export function requestAttachment(
...
+    }
+  } = client;
+  return fetch(`${remote}/permissions`, {
```
Natim: Shouldn't we add that to kinto-http.js ?

Natim:
f+

scripts/client.js
n1k0: This is a flow type.

scripts/client.js
n1k0: Yes we should, just not now.

n1k0:
We'll badly need to add functional tests there.

scripts/sagas/session.js
```diff
@@ -47,12 +47,34 @@ export function* listBuckets(getState, action) {
     const responses = yield call([client, client.batch], (batch) => {
       for (const {id} of data) {
         batch.bucket(id).listCollections();
+        // XXX future: query the /permissions endpoint to retrieve the other
+        // collections we may have access to bcause they're shared with us
```
leplatrem: typo because

scripts/sagas/session.js
```diff
@@ -47,12 +47,34 @@ export function* listBuckets(getState, action) {
...
+    // the /permissions endpoint
+    for (const permission of permissions.data) {
+      // Add any missing bucket to the current list
+      let bucket = buckets.find(x => x.id === permission.bucket_id);
+      if (!bucket) {
+        bucket = {id: permission.bucket_id, collections: []};
+        buckets.push(bucket);
+      }
+      // Add any missing collection to the current bucket collections list
+      if ("collection_id" in permission) {
```
leplatrem: Add a comment that it covers records and collections

leplatrem:
Nicely done!

This will also work for individual records that are shared in private collections :) Also because the server will return a partial records list when querying a collection where only a couple of records are shared :)

One of the next steps that comes into my mind (further issue/pr) would be to show edit button on buckets/collection/records only if write is granted! Awesome :)

scripts/sagas/session.js
```diff
@@ -47,12 +47,34 @@ export function* listBuckets(getState, action) {
...
         batch.bucket(id).listCollections();
+        // XXX future: query the /permissions endpoint to retrieve the other
+        // collections we may have access to bcause they're shared with us
       }
     });
     const buckets = data.map((bucket, index) => {
       const collections = responses[index].body.data;
       return {id: bucket.id, collections};
     });
+    const permissions = yield call(requestPermissions);
```
leplatrem: Having a `if` here for a (capability/protocol version) check wouldn't be too much hassle actually

n1k0: Well, indeed, though we should probably have a capability exposed in the first place :)

leplatrem: If we make it into the core, the check will have to be made on `http_protocol` version number

n1k0:
> One of the next steps that comes into my mind (further issue/pr) would be to show edit button on buckets/collection/records only if write is granted!

Please note that even if we display these, if you don't have write permissions you'll get a warning rendered and disabled edit forms.

scripts/sagas/session.js
n1k0: Ah, right.

n1k0:
@leplatrem r=? for https://github.com/Kinto/kinto-admin/pull/161/commits/5a167983e98334c0d485933b1a464418737d0729

n1k0:
This has been pair-reviewed with @magopian and @leplatrem.

leplatrem:
r+!

PR #162, opened 2016-06-28
Split dev dependencies in package.json

leplatrem:


n1k0:
See previous discussion about this https://github.com/Kinto/kinto-admin/pull/137#discussion_r66522966

I suggest we reintroduce that split the day we publish a package actually allowing to leverage any js API, eg. for plugins.

PR #163, opened 2016-06-29
Fixes #155: Add recently used servers history selector

### LINKED ISSUE #155: Add a selector for recently used servers
Having the moz dev server by default is nice for demo, but may gets annoying when the admin is used on a daily basis with other servers. We should add a selector for recently used servers so that users can quickly pick one.

We should store the list of recently used servers in localStorage. We should only add a server to the history when it's been successfully used.

Optionally, we may consider storing a couple of server/username (password *not* included).

n1k0:
Refs #155. this patch:

- exposes a UI to select a previously used server;
- prefill the latest used server in the server field. 

Deployed to gh-pages.


n1k0:
Pair and pair-reviewed with @magopian

PR #174, opened 2016-07-06
Fix #173: Store user entered server URL in servers history.

### LINKED ISSUE #173: Do not store the url property of the serverInfo in history, rather the one entered by the user
The `url` property of the `serverInfo` served at `/v1/` should not be relied upon as it may contain a private (*"secret"*) address we do not want to expose in the history dropdown.

n1k0:
r=? @Natim @leplatrem


leplatrem:
r+

PR #175, opened 2016-07-07
Fix broken listBuckets operations.

n1k0:
This fixes a breakage on current master where the list of buckets and collections are not properly refreshed after a transactional operation is performed. This has been introduced by #174.


Natim:
r+

PR #179, opened 2016-07-19
Upgrade kinto-http.js to 2.1.1.

n1k0:
This patch leverages the new kinto-http.js attachment API. Patch deployed to gh-pages.

r=? @leplatrem

### DIFF
--- a/scripts/sagas/collection.js
+++ b/scripts/sagas/collection.js
@@ -1,9 +1,8 @@
 import { push as updatePath } from "react-router-redux";
 import { call, put } from "redux-saga/effects";
 import { v4 as uuid } from "uuid";
-import { createFormData } from "../utils";
 
-import { getClient, requestAttachment } from "../client";
+import { getClient } from "../client";
 import { notifySuccess, notifyError } from "../actions/notifications";
 import { resetRecord } from "../actions/record";
 import * as actions from "../actions/collection";
@@ -27,8 +26,9 @@ function getCollection(bid, cid) {
 export function* deleteAttachment(getState, action) {
   const {bid, cid, rid} = action;
   try {
+    const coll = getCollection(bid, cid);
     yield put(actions.collectionBusy(true));
-    yield call(requestAttachment, bid, cid, rid, {method: "delete"});
+    yield call([coll, coll.removeAttachment], rid);
     yield put(updatePath(`/buckets/${bid}/collections/${cid}/edit/${rid}`));
     yield put(notifySuccess("Attachment deleted."));
   } catch(error) {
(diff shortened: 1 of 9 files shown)


scripts/sagas/collection.js
```diff
@@ -55,19 +55,16 @@ export function* listRecords(getState, action) {
...
-    if (shouldProcessAttachment(session.serverInfo, record)) {
-      const rid = yield call(uuid);
-      const formData = yield call(createFormData, record);
-      yield call(requestAttachment, bid, cid, rid, {
-        method: "post",
-        body: formData
-      });
+    if (shouldProcessAttachment(session.serverInfo, rawRecord)) {
+      const id = yield call(uuid);
+      const {__attachment__: dataURL, ...record} = rawRecord;
```
leplatrem: I think we should add a comment to remind the reader where does this `__attachment__` comes from :)

scripts/sagas/collection.js
```diff
@@ -55,19 +55,16 @@ export function* listRecords(getState, action) {
...
+      yield call([coll, coll.addAttachment], dataURL, {...record, id});
     } else {
-      yield call([coll, coll.createRecord], record);
+      yield call([coll, coll.createRecord], rawRecord);
```
leplatrem: And especially here make it clear that there is `__attachment__` attribute being sent to the server

test/sagas/collection_test.js
```diff
@@ -332,18 +329,9 @@ describe("collection sagas", () => {
...
-
         it("should update the record with its attachment", () => {
-          const fakeFormData = {fake: true};
-          expect(updateRecord.next(fakeFormData).value)
-            .eql(call(requestAttachment, "bucket", "collection", 1, {
-              method: "post",
-              body: fakeFormData,
-            }));
+          expect(updateRecord.next().value)
+            .eql(call([collection, collection.addAttachment], {}, record));
```
leplatrem: isn't it supposed to be a `dataURL` as first argument ?

scripts/sagas/collection.js
leplatrem: (had to read the `shouldProcessAttachment` function

leplatrem:
r+, some comments would help fresh readers though

test/sagas/collection_test.js
n1k0: It's because we define fake attachments as `{}`. I can update that though.

scripts/sagas/collection.js
n1k0: I can probably update the action payload to separate between record data and attachment to ease reading the code.

PR #181, opened 2016-07-19
Fix #180: Check for the permissions_endpoint capability.

### LINKED ISSUE #180: Check for the permissions_endpoint capability before requesting it
A `permissions_endpoint` capability will inform us if we can query the new permission endpoint, see https://github.com/Kinto/kinto/pull/724

n1k0:
Refs #180 & waiting for https://github.com/Kinto/kinto/pull/724

r=? @Natim


Natim:
r+

Issue #189, opened 2016-08-19
Font issue


Natim:
When using the ``kinto-admin build -d kinto_admin/static/`` script to package the kinto-admin it seems that the font are not loaded correctly.

See: https://github.com/mozilla-services/kinto-dist/blob/master/Makefile#L44-L48

![kinto-writer-font](https://cloud.githubusercontent.com/assets/229453/17806527/6d9a2f3e-6605-11e6-867b-635a8aa6ef4e.png)

n1k0:
You need to set the `KINTO_ADMIN_PUBLIC_PATH` env var to the exposed absolute path url of the assets folder before running `npm run dist`, as you can see in `package.json` commands.

n1k0:
See #190 for docs patch.

n1k0:
Fixed by #189. Underlying issue will be fully addressed by #192.

PR #190, opened 2016-08-19
Refs #189: Document the KINTO_ADMIN_PUBLIC_PATH env var.

n1k0:
refs #189. r=? @Natim

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,6 +189,18 @@ For instance:
 If you use `target.merkle.tree.file.name` it will render the string
 `foobar` and `target.proof.hash` will render `abcd`.
 
+### Assets path url
+
+By default this package assumes the static admin will be hosted at the root of a given domain; if you plan on hosting it under a given sub path, you need to build the admin setting the `KINTO_ADMIN_PUBLIC_PATH` env var, specifying the absolute root URL path where the static asset files are obtainable.
+
+For example, if you plan on hosting the admin at `https://mydomain.tld/kinto-admin/`, you need to build it like this:
+
+```
+$ KINTO_ADMIN_PUBLIC_PATH=/kinto-admin/ npm run dist
+```
+
+Note: Unfortunately, the `kinto-admin build` CLI command doesn't support this feature.
+
 ## License
 
 Apache Licence Version 2.0


README.md
```diff
@@ -189,6 +189,18 @@ For instance:
...
+
+By default this package assumes the static admin will be hosted at the root of a given domain; if you plan on hosting it under a given sub path, you need to build the admin setting the `KINTO_ADMIN_PUBLIC_PATH` env var, specifying the absolute root URL path where the static asset files are obtainable.
+
+For example, if you plan on hosting the admin at `https://mydomain.tld/kinto-admin/`, you need to build it like this:
+
+```
+$ KINTO_ADMIN_PUBLIC_PATH=/kinto-admin/ npm run dist
+```
+
+Note: Unfortunately, the `kinto-admin build` CLI command doesn't support this feature.
```
Natim: Could we fix this by adding an option to it?

n1k0: Not easily as that would mean exposing the whole building logic to the package executable; right now the `kinto-admin` executable simply copies already built files, so that's simple, fast and efficient. What you suggest would imply on-demand building the admin solely to tweak some path variable, which sounds slightly overkill methinks.

Natim: Is it possible to change the PUBLIC_PATH in the generated files?

n1k0: Not easily, that's why for this purpose right now the solution is to build the admin itself setting the `KINTO_ADMIN_PUBLIC_PATH` var. Note that this is mostly matter of working from the repo instead of the installed package, running:

```
$ KINTO_ADMIN_PUBLIC_PATH=/plop/ npm run dist 
```

Instead of

```
$ kinto-admin build
```

I understand and agree that this is a little less convenient, though moving the package executable to on-demand building is a big and schedule-risky task really.

Note: Depending on the future of plugins, we could also consider this opportunity to offer on-demand building of the admin with selected plugins as well, so that would be killing two (large) birds with one stone. I wasn't expecting to work on it this soon, but that may be the right moment to start the move. 

Now the embarrassing question: would you want to work on this? :)

n1k0: Also note that if you can find any smart simple way to post-update the path the CSS have been built against, I take it too :) (no, sed is not an option)

Natim: It just means the ``kinto-admin build`` CLI tool is not useful and you cannot use ``npm run dist`` from the ``npm install kinto``

Natim: > Also note that if you can find any smart simple way to post-update the path the CSS have been built against, I take it too :) (no, sed is not an option) 

Yes it what I was thinking about. I will have a look at it.

n1k0: I've filed #192 about this as this is something we'll eventually want anyway.

Natim:
Ok r+ then :)

PR #191, opened 2016-08-19
Fix #188: Add a delete button in the record edition view.

### LINKED ISSUE #188: Remove an entry when there is an attachment.
I didn't see where I could delete a record:

I think a button in the Edit part could be nice:

capture du 2016-08-19 11-57-07

Apparently the button in the list is not accessible when there is an attachment:

vue liste

n1k0:
Refs #188. This also fixes the CSS issues on Firefox hiding the action buttons in the records list. Deployed to gh-pages.

r=? @Natim @leplatrem

### DIFF
--- a/css/styles.css
+++ b/css/styles.css
@@ -124,8 +124,8 @@ table.record-list tr.conflicting {
   );
 }
 
-table.record-list td {
-  max-width: 10vw;
+table.record-list td:not(.actions) {
+  max-width: 8vw;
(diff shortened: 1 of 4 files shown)


Natim:
r+

n1k0:
Released in [v1.2.1](https://github.com/Kinto/kinto-admin/releases/tag/v1.2.1).

PR #195, opened 2016-08-23
Fix bulk creation error when no attachment is provided.

n1k0:
This fixes an issue where when bulk creating records when the attachments capability is enabled on the server, records without any attachment specified would make the whole operation fail.

Here we ensure that an attachment exists for each record, falling back to the standard API if none is provided.

### DIFF
--- a/scripts/sagas/collection.js
+++ b/scripts/sagas/collection.js
@@ -114,8 +114,12 @@ export function* bulkCreateRecords(getState, action) {
       // XXX We should perform a batch request here
       for (const rawRecord of records) {
         // Note: data urls are attached to the __attachment__ record property
-        const {__attachment__: dataURL, ...record} = rawRecord;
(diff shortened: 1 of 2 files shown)


leplatrem:
r+

scripts/sagas/collection.js
```diff
@@ -114,8 +114,12 @@ export function* bulkCreateRecords(getState, action) {
       // XXX We should perform a batch request here
       for (const rawRecord of records) {
         // Note: data urls are attached to the __attachment__ record property
-        const {__attachment__: dataURL, ...record} = rawRecord;
-        yield call([coll, coll.addAttachment], dataURL, record);
+        const {__attachment__: attachement, ...record} = rawRecord;
+        if (attachement) {
```
glasserc: `attachment`?

n1k0: MEGAWOOPS.

PR #201, opened 2016-08-25
Fix #200: Reset the list of paginated records when order changes.

### LINKED ISSUE #200: Loaded records list chunks are not reordered when a sort button in clicked
cc @Natim

n1k0:
This fixes an issue (#200) where a paginated list of records already loaded isn't reordered when the user clicks on a new sort button.

Here I'm just resetting the list of records everytime a new sort parameter is used.

Also added a few missing test cases for pagination in the reducers. Patch deployed to gh-pages.

### DIFF
--- a/scripts/reducers/collection.js
+++ b/scripts/reducers/collection.js
@@ -79,7 +79,10 @@ export function collection(
       };
     }
     case COLLECTION_RECORDS_REQUEST: {
-      return {...state, sort: action.sort, recordsLoaded: false};
+      // If a new sort filter is used, purge the previous records list and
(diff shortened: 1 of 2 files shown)


Natim:
r+

leplatrem:
r+ !

There is one idea (*I let open the issue if relevant*) : if the list if fully loaded (no more pages), we could sort in memory to avoid going to the server

n1k0:
> if the list if fully loaded (no more pages), we could sort in memory to avoid going to the server

Definitely a neat idea, though let's iterate on this. Landing.

PR #202, opened 2016-08-26
CRUD groups

leplatrem:
* [x] URL / forms / cabling
* [x] Write some tests
* [x] Type flow stuff


n1k0:
First review pass done, it just looks great. Kudos!

leplatrem:
Ready to merge/review :) `/me goes to grab lunch`

test/sagas/bucket_test.js
```diff
@@ -18,6 +19,12 @@ const collectionData = {
   displayFields: [],
 };
 
+const groupData = {
+  id: "group",
+  members: ["tartenpion", "account:abc"]
```
n1k0: major nit: *tartempion* :p

scripts/components/CollectionHistory.js
```diff
@@ -52,7 +52,7 @@ class HistoryRow extends Component {
   }
 }
 
-function HistoryTable(props) {
+export function HistoryTable(props) {
```
n1k0: This is certainly generic enough to live in its own module.

leplatrem: We did it in the other pull-request

n1k0:
I've now read this huge piece of diff twice and couldn't spot any major mistake, and running the patch locally works and provides the expected features :)

The only bit I'm slightly uncomfortable with is the new loadRoute saga signature but couldn't find any better way without a big scary refactoring involved, so I think we can live with it for now.

r+

PR #215, opened 2016-08-29
Fix #212: Add breadcrumb trails.

### LINKED ISSUE #212: Add breadcrumb trails
Now we have tabs (many of them), navigation can get confusing eg. when you switch from a bucket to a collection. It may be worth adding [breadcrum trails](https://en.wikipedia.org/wiki/Breadcrumb_(navigation)) so users know better where they currently are in the resource hierarchy.

n1k0:
Refs #212. This adds breadcrumb navigation to every page in the admin. Deployed to gh-pages.

r=? @Natim @glasserc


Natim:
Oh yeah :)

Natim:
Looks really good :)

.flowconfig
```diff
@@ -7,6 +7,7 @@
 
 [libs]
 ./node_modules/flow-interfaces/interfaces/
+./node_modules/iflow-react-router/index.js.flow
```
n1k0: So I have to use this because the one provided in flow-interfaces is obsolete.

Natim:
I think when you click on the bucket name and you are in the collections page you should go to its property page instead of the collections one but maybe it is just me.

n1k0:
I debated that and I don't know really. Ideally we'd have a page listing all the buckets, which is something we could eventually implement in another PR.

Natim:
> Ideally we'd have a page listing all the buckets

I though this was already what the sidebar was doing.

n1k0:
The probability for users to want accessing the properties instead of the collections the bucket contains seems slightly more unlikely to me, but that may just be me as well so I don't know really. If you rea

n1k0:
> I though this was already what the sidebar was doing.

You thought well, but this is not a page, it's a sidebar, you know.

Natim:
Ok so let's land it as this and we can iterate later :)

n1k0:
You know what you seem to have opinions on this and i don't so I'm just gonna implement whatever you decide.

n1k0:
Note: I've tried to factor common component props into some higher order `Route` component but it didn't go well. Seems like react-router doesn't actually execute the Route components, see http://jsbin.com/huziho/4/edit?html,js,output

Oh well, landing this as is.

PR #196, opened 2016-08-23
Add views for browsing collections history

n1k0:
Feedback=? @leplatrem

PR #216, opened 2016-08-29
Fix #214: Remove collection contextual menus in the sidebar.

### LINKED ISSUE #214: Remove the collection contextual menu in the sidebar
It really gets in the way. We should return to the previous behavior where clicking the cog icon leads to the collection properties page.

n1k0:
Refs #214. As discussed with @leplatrem, the contextual action menu for collection entries in the sidebar is painful to use and should be removed.

Note: this mostly reverts changes done in #196 for the sidebar.

r=? @Natim

### DIFF
--- a/css/styles.css
+++ b/css/styles.css
@@ -247,22 +247,6 @@ table.record-list td.load-more {
   color: #333;
 }
 
-.gear-menu {
-  float: right;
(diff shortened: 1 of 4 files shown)


Natim:
r+

PR #217, opened 2016-08-29
Fixes #213: Fix redirect URL handling after successful authentication.

### LINKED ISSUE #213: The authentication required notification should be cleared once authenticated
This is annoying as hell.

n1k0:
This restores the expected behavior when loading an route requiring being authenticated when you're currently not. That also fixes the issue described in #213 where the *Authentication required* notification wouldn't go away after successfully authenticating.

Deployed to gh-pages.

QA:

1. Head to https://kinto.github.io/kinto-admin/#/buckets/Foo/collections/Bar
1.  You should be redirected to the authentication form with an *Authentication required* notification displayed
1. Authenticate using test:test (use the default dev kinto server url)
1. You should be redirected to the collection list view, and the notification should be gone. 

r=? @Natim @glasserc

### DIFF
--- a/scripts/reducers/session.js
+++ b/scripts/reducers/session.js
@@ -14,6 +14,7 @@ import {
 const DEFAULT: Session = {
   busy: false,
   authenticated: false,
+  authPending: false,
   server: null,
   credentials: {},
   buckets: [],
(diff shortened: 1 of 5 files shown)


n1k0:
Superseded by #218.

PR #218, opened 2016-08-30
Fix #213: Fix redirect after auth flow.

### LINKED ISSUE #213: The authentication required notification should be cleared once authenticated

n1k0:
This supersedes #217 with a better implementation. pair=@Natim


n1k0:
pair=@Natim

PR #221, opened 2016-09-01
Refactored notification API and user facing error messages.

### LINKED ISSUE #220: No error when creating a bucket that already exists
Same goes with collections. We should prevent the transaction and inform users instead.

n1k0:
Also fixes #220. This exposes user friendly error messages in case server requests fail, while allowing to access technical details with a convenient button.

r=? @Natim

### DIFF
--- a/scripts/actions/session.js
+++ b/scripts/actions/session.js
@@ -64,6 +64,6 @@ export function navigateToExternalAuth(authFormData: Object): ?Action {
     const redirect = encodeURIComponent(`${origin}${pathname}#/auth/${payload}/`);
     document.location.href = `${server}/fxa-oauth/login?redirect=${redirect}`;
   } catch(error) {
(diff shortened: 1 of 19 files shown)


n1k0:
Deployed to gh-pages. Try creating a new bucket reusing an existing one's name.

scripts/actions/session.js
```diff
@@ -64,6 +64,6 @@ export function navigateToExternalAuth(authFormData: Object): ?Action {
     const redirect = encodeURIComponent(`${origin}${pathname}#/auth/${payload}/`);
     document.location.href = `${server}/fxa-oauth/login?redirect=${redirect}`;
   } catch(error) {
-    return notifyError(error);
+    return notifyError("Couldn,'t redirect to authentication endpoint.", error);
```
glasserc: `Couldn't`

PR #226, opened 2016-09-02
Fix #225: Ensure reauth on session restore.

### LINKED ISSUE #225: Force reauth on session restore
We shouldn't blindly reuse data stored in a saved session, eg. the list of buckets as they might have been heavily updated meanwhile.

We should ensure fully reauthenticating when loading previsouly saved session data.

n1k0:
Refs #225.

> We shouldn't blindly reuse data stored in a saved session, eg. the list of buckets as they might have been heavily updated meanwhile.
>
> We should ensure fully reauthenticating when loading previsouly saved session data.


n1k0:
r=@me

PR #230, opened 2016-09-06
Reorganized repository root.

n1k0:
This creates two new folders at the root of the repository:

- `html/` which stores the dev and prod index HTML files;
- `server/` which stores both the dev and test server code.


n1k0:
I'm not gonna lie, I'm landing this.

PR #231, opened 2016-09-07
Fix some technical debt

n1k0:
Refs #204 and addresses a few other bits of technical debt.

- [x] Removed unused `ROUTE_LOAD_SUCCESS` action
- [x] Rename `Bucket` type to `BucketState`
- [x] Rename `Collection` type to `CollectionState`
- [x] Rename `Session` type to `SessionState`
- [x] Rename `SessionServerInfo` to `ServerInfo`
- [x] Update default collection records route to `/collection/:cid/records`
- [x] Isolate collection data in collection state
- [x] Drop unused collection state `bucket` property
- [x] Rename collection state `name` property to `id`
- [x] Rename bucket state `name` property to `id`

PR #235, opened 2016-09-07
234 confusing sidebar

Natim:
While browsing the sidebar code I discovered two little glitch.


n1k0:
This is being addressed by #231 :)

I'm not gonna land this patch because the user bucket detection is currently handled from the reducer, so this code is actually never executed; it's a leftover from a time where it was handled in the sidebar component.

PR #240, opened 2016-09-08
Make sure to check all the permissions in the kinto-admin. (Fixes #186)

### LINKED ISSUE #186: Record update permissions are not recognized for owner of bucket/collection.
kinto-admin doesn't seem to recognize update permissions on records for owners of buckets/collections containing those.

#### Steps to reproduce ####

1. Create a bucket `b` with user `A`
2. Create a collection `c` with the same user `A`.
3. Create a record `r` with a user `B`.
4. User `A` won't be able to update the record via kinto-admin even though it can via http API.

####  Additional info ####

When a record is created in a collection by a certain user `B`, that record will be of the type:
```
{ 
  "data": { ... }
  "permissions": {
    "write": [
      "B"
    ]
  }
}
```
and there is nothing written on the record about the owner of the collection,

However, after updating via HTTP API with the owner of the collection, user `A` is added to the `write` array, as in `permissions: { write: [ "A", "B" ] }`. After this, kinto-admin will start allowing updates from user `A`, as per expected initial behavior.

Natim:
Refs #186 

The function that gives permissions to create or edit buckets/collections/groups/records currently don't take care of the permission inheritance. This PR take care of fixing this. You can see #186 for more details about it but I added tests.

r? @n1k0


n1k0:
Would you mind writing a few sentences the situation this patch is trying to fix? Thanks! :)

Natim:
The function that give permissions to create or edit buckets/collections/groups/records currently don't take care of the permission inheritance. This PR take care of fixing this. You can see #186 for more details about it but I added tests.

n1k0:
I fixed the component tests, but it's not entirely satisfying as we're passing way too many generic props to components. Here I'm realizing we should precompute what's allowed to do in the containers (eg. through dedicated selectors) and only propagate these value as component props.

I'll play around with this idea.

src/components/collection/CollectionForm.js
```diff
@@ -228,7 +228,7 @@ export default class CollectionForm extends Component {
     if (creation) {
       return canCreateCollection(session, bucket);
     } else {
-      return canEditCollection(session, collection);
+      return canEditCollection(session, collection, bucket);
```
Natim: Change arguments order to be ``session, bucket, collection, record``

src/permission.js
```diff
@@ -41,32 +41,57 @@ export function can(session: SessionState): Object {
...
+  return canSession.write(group) || canSession.write(bucket);
 }
 
-export function canCreateRecord(session: SessionState, collection: CollectionState): boolean {
+export function canCreateRecord(session: SessionState, collection: CollectionState, bucket: BucketState): boolean {
   const canSession = can(session);
-  return canSession.write(collection) || canSession.createRecord(collection);
+
+  // You can edit if you can write in the collection or create a record in it.
+  let canCreate = canSession.write(collection) || canSession.createRecord(collection);
```
Natim: Use `const` and return to new line after each `||` with a comment in between.

src/components/group/GroupForm.js
```diff
@@ -93,7 +93,7 @@ export default class GroupForm extends Component {
     const {gid, session, bucket, group, formData, deleteGroup} = this.props;
     const creation = !formData;
     const hasWriteAccess = creation ? canCreateGroup(session, bucket)
-                                    : canEditGroup(session, group);
+                                    : canEditGroup(session, group, bucket);
```
n1k0: A group depends on a bucketThe parent of a group being a bucket, I suggest the `canEditGroup(session, bucket, group)` instead

src/permission.js
```diff
@@ -41,32 +41,51 @@ export function can(session: SessionState): Object {
...
 
 export function canCreateGroup(session: SessionState, bucket: BucketState): boolean {
+  // You can create a group if you can write in the bucket or create a group in it.
   const canSession = can(session);
   return canSession.write(bucket) || canSession.createGroup(bucket);
 }
 
-export function canEditGroup(session: SessionState, group: GroupState): boolean {
-  return can(session).write(group);
+export function canEditGroup(session: SessionState, bucket: BucketState, group: GroupState): boolean {
```
n1k0: Which is funny because that's the actual signature of the method :D

src/permission.js
```diff
@@ -41,32 +41,51 @@ export function can(session: SessionState): Object {
...
-  return can(session).write(record);
+export function canEditRecord(session: SessionState, bucket: BucketState, collection: CollectionState, record: RecordState): boolean {
+  const canSession = can(session);
+
+         // You can edit if you can write in the record
+  return canSession.write(record) ||
+         // You can edit if you can write in the collection
+         canSession.write(collection) ||
+         // You can edit if you can write in the bucket
+         canSession.write(bucket);
```
n1k0: Alternatively:

```js
return [
  canSession.write(record),
  canSession.write(collection),
  canSession.write(bucket),
].some(condition => condition);
```

n1k0:
r+ with nits :)

test/permission_test.js
```diff
@@ -104,15 +106,81 @@ describe("canEditCollection", () => {
...
+});
+
+describe("canEditGroup", () => {
+  const session = {authenticated: true, serverInfo: {user: {id: 1}}};
+
+  it("should check if a group can be edited", () => {
+    const bucket = {permissions: {}};
+    const group = {permissions: {write: [1]}};
+
+    expect(canEditGroup(session, group, bucket)).eql(true);
```
Natim: bucket, group

test/permission_test.js
```diff
@@ -104,15 +106,81 @@ describe("canEditCollection", () => {
...
+  });
+
+  it("should check if a bucket can be edited", () => {
+    const bucket = {permissions: {write: [1]}};
+    const group = {permissions: {}};
+
+    expect(canEditGroup(session, group, bucket)).eql(true);
```
Natim: bucket, group

test/permission_test.js
```diff
@@ -104,15 +106,81 @@ describe("canEditCollection", () => {
...
+  });
+
+  it("should check if a group cannot be edited", () => {
+    const bucket = {permissions: {write: ["chucknorris"]}};
+    const group = {permissions: {write: ["chucknorris"]}};
+
+    expect(canEditGroup(session, group, bucket)).eql(false);
```
Natim: bucket, group

src/permission.js
```diff
@@ -41,32 +41,46 @@ export function can(session: SessionState): Object {
 }
 
 export function canEditBucket(session: SessionState, bucket: BucketState): boolean {
+  // You can edit a bucket if you can write in it.
   return can(session).write(bucket);
 }
 
 export function canCreateCollection(session: SessionState, bucket: BucketState): boolean {
+  // You can createa a collection if you can write in the bucket or create a collection in it.
```
n1k0: nit: *create a*

src/components/record/RecordForm.js
```diff
@@ -125,11 +125,11 @@ export default class RecordForm extends Component {
...
 
   get allowEditing() {
-    const {session, collection, record} = this.props;
+    const {session, bucket, collection, record} = this.props;
     if (record) {
-      return canEditRecord(session, record);
+      return canEditRecord(session, bucket, collection, record);
     } else {
-      return canCreateRecord(session, collection);
+      return canCreateRecord(session, bucket, collection);
```
n1k0: We may want to check for a record being provided in the `canCreateRecord` function instead.

Natim: I didn't get that, what do you mean?

n1k0: Bah, misred the function name, sratch that. (will delete these two comments after you've been notified about them)

src/permission.js
```diff
@@ -41,32 +41,46 @@ export function can(session: SessionState): Object {
...
   const canSession = can(session);
   return canSession.write(bucket) || canSession.createCollection(bucket);
 }
 
-export function canEditCollection(session: SessionState, collection: CollectionState): boolean {
-  return can(session).write(collection);
+export function canEditCollection(session: SessionState, bucket: BucketState, collection: CollectionState): boolean {
+  // You can create a group if you can write in the bucket or create a group in it.
```
n1k0: Why are we mentioning group creation here?

src/permission.js
```diff
@@ -41,32 +41,46 @@ export function can(session: SessionState): Object {
...
+  return [bucket, group].some(can(session).write);
 }
 
-export function canCreateRecord(session: SessionState, collection: CollectionState): boolean {
+export function canCreateRecord(session: SessionState, bucket: BucketState, collection: CollectionState): boolean {
   const canSession = can(session);
-  return canSession.write(collection) || canSession.createRecord(collection);
+
+  return [
+    // You can edit if you can write in the collection.
```
n1k0: s/edit/create

test/permission_test.js
```diff
@@ -121,45 +189,77 @@ describe("canCreateRecord", () => {
...
+    });
   });
 
-  it("should check if a record cannot be edited", () => {
-    const record = {permissions: {write: ["chucknorris"]}};
+  describe("bucket write permission", () => {
+    it("should check if a record can be edited", () => {
+      const bucket = {permissions: {write: [1]}};
+      const collection = {permissions: {write: ["chucknorris"]}};
+      const record = {permissions: {write: ["chucknorris"]}};
```
n1k0: General remark in this test suite: for better understandability, it may be worth switching from `chucknorris` and `1` to `current-user` and `someone-else`. WDYT?

Natim: For sure :)

PR #241, opened 2016-09-08
Fix #239: Refactor route resources loading flow.

### LINKED ISSUE #239: Refactor route loading sagas
We should really clean this part of the code.

n1k0:


glasserc:
Sometimes it's hard for me to review this kind of PR because I don't really understand it. I'm going to ask some questions to try to get a better sense of what's happening. These aren't critiques, just an attempt to follow along with what you're doing so that I can review stuff in the future if I need to.

What is a "route resource"? It seems like a remote resource, meaning either a bucket, a group, a collection, or a record. Is that correct?

What is being refactored about it? It seems like you're pulling together these resources and DRYing out some logic about deciding whether you are busy or not. Is that correct?

n1k0:
> What is a "route resource"? It seems like a remote resource, meaning either a bucket, a group, a collection, or a record. Is that correct?

Yes, basically a unique server resource mapped to a URL parameter. We fetch these automatically on each route change in a batch HTTP request.

> What is being refactored about it? It seems like you're pulling together these resources and DRYing out some logic about deciding whether you are busy or not. Is that correct?

Yes.

PR #243, opened 2016-09-09
Fix #242: Fix broken sort links.

### LINKED ISSUE #242: Broken records sort links
We've removed the `sort` collection state property while it still makes sense to know about the *current* sort filter to be used.

n1k0:
Refs #242. r=? @Natim


Natim:
LGTM

Natim:
r+

PR #246, opened 2016-09-14
Fix #245: Empty buckets list when saving session.

### LINKED ISSUE #245: Don't store sidebar state in saved session data
When loading the admin restoring a previously saved session, the state of the sidebar (the list of buckets and collections) is restored as well, while these data might be quite heavily outdated; while the new list is loaded, the sidebar renders links to possibly deleted buckets and collection, and if users click on these, they'll get 403 or 404 errors.

We should ensure clearing these specific data before saving a full session dump to localStorage, so we never render them after a session is reloaded.

n1k0:
r=? @Natim

### DIFF
--- a/src/store/localStore.js
+++ b/src/store/localStore.js
@@ -41,7 +41,10 @@ export function loadSession(): ?Object {
 }
 
 export function saveSession(sessionState: SessionState) {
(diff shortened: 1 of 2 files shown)


Natim:
r+

PR #251, opened 2016-10-04
Show relative time in history entries (fixes #207)

### LINKED ISSUE #207: Display relative time in history entries
Use a library like *moment.js* to show `3 minutes ago` or `6 months ago` instead `2016-01-13T14:30`

leplatrem:
capture d ecran de 2016-10-04 11-00-19

@n1k0 r?

### DIFF
--- a/src/components/HistoryTable.js
+++ b/src/components/HistoryTable.js
@@ -27,7 +27,7 @@ class HistoryRow extends Component {
       <tbody>
         <tr>
           <td>
-            <span title={date}>{timeago(date)}</span>
+            <span title={date}>{timeago(date + "Z")}</span>
           </td>
           <td>{action}</td>
           <td>{resource_name}</td>
--- a/src/components/HistoryTable.js
+++ b/src/components/HistoryTable.js
@@ -23,11 +23,12 @@ class HistoryRow extends Component {
       target,
       user_id,
     } = entry;
+    const tzDate = date + " UTC";
     return (
       <tbody>
         <tr>
           <td>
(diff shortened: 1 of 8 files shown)


src/components/HistoryTable.js
```diff
@@ -22,10 +23,13 @@ class HistoryRow extends Component {
       target,
       user_id,
     } = entry;
+    const tzDate = date + " UTC";
     return (
       <tbody>
         <tr>
-          <td>{date}</td>
+          <td>
+            <span title={tzDate}>{timeago().format(new Date(tzDate))}</span>
```
n1k0: Would be nice to move this to a function in `utils`, wdyt?

src/utils.js
```diff
@@ -2,6 +2,7 @@
 
 import type { RecordData } from "./types";
 import React from "react";
+import { default as _timeago } from "timeago.js";
```
n1k0:
```js
import _timeago from "timeago.js";  
```

n1k0:
Flow says:

> src/utils.js:5
> ```
> import { default as _timeago } from "timeago.js";
>                                     ^^^^^^^^^^^^ timeago.js. Required module not found
```

I suspect we have to write the flow type definition for this library. This is to be done in `interface/external_modules`; have a look at how it's done for other external packages we use in the project :)

leplatrem:
Should we use it as well in the records list ?

n1k0:
> Should we use it as well in the records list ?

Yes! Putting the date in a tooltip too :)

leplatrem:
Done

n1k0:
r+

PR #255, opened 2016-10-04
Concurrency headers on create/update/delete

leplatrem:
* [x] Safe creations
* [x] Safe deletions
* [x] Safe updates
* [x] Rework tests

### DIFF
--- a/src/components/bucket/BucketEdit.js
+++ b/src/components/bucket/BucketEdit.js
@@ -7,11 +7,10 @@ import BucketTabs from "./BucketTabs";
 
 export default class BucketEdit extends Component {
   deleteBucket = (bid) => {
-    const {deleteBucket, bucket} = this.props;
-    const {last_modified} = bucket;
+    const {deleteBucket} = this.props;
     if (confirm("This will delete the bucket and all the collections and " +
                 "records it contains. Are you sure?")) {
-      deleteBucket(bid, last_modified);
+      deleteBucket(bid);
     }
   };
 
--- a/src/components/bucket/BucketEdit.js
+++ b/src/components/bucket/BucketEdit.js
@@ -16,11 +16,9 @@ export default class BucketEdit extends Component {
   };
 
   onSubmit = (formData) => {
-    const {params, updateBucket, bucket} = this.props;
+    const {params, updateBucket} = this.props;
     const {bid} = params;
-    const {last_modified} = bucket;
-    const savedBucket = {...formData, last_modified};
-    updateBucket(bid, savedBucket);
+    updateBucket(bid, formData);
   }
 
   render() {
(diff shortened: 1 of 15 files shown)


src/reducers/bucket.js
```diff
@@ -20,6 +20,7 @@ import { omit } from "../utils";
 const INITIAL_STATE: BucketState = {
   busy: false,
   id: null,
+  last_modified: null,
```
n1k0: How about we keep it in the `data` object above instead?

src/reducers/bucket.js
```diff
@@ -40,11 +41,12 @@ function load(state: BucketState, bucket: BucketResource): BucketState {
...
   }
   const {data, permissions} = bucket;
-  const {id} = data;
+  const {id, last_modified} = data;
   return {
     ...state,
     busy: false,
     id,
+    last_modified,
     data: omit(data, ["id", "last_modified"]),
```
n1k0: Maybe we should not omit `last_modified` here?

src/components/bucket/BucketEdit.js
```diff
@@ -7,17 +7,20 @@ import BucketTabs from "./BucketTabs";
 
 export default class BucketEdit extends Component {
   deleteBucket = (bid) => {
-    const {deleteBucket} = this.props;
+    const {deleteBucket, bucket} = this.props;
+    const {last_modified} = bucket;
     if (confirm("This will delete the bucket and all the collections and " +
                 "records it contains. Are you sure?")) {
-      deleteBucket(bid);
+      deleteBucket(bid, last_modified);
```
n1k0: I'm uncomfortable doing this in a view component; I suspect we could inspect the resource data state from the store in a saga to infer what should be sent to the server?

src/sagas/bucket.js
```diff
@@ -50,11 +50,11 @@ export function* updateBucket(getState, action) {
 }
 
 export function* deleteBucket(getState, action) {
-  const { bid } = action;
+  const { bid, last_modified } = action;
```
n1k0: Note that from here, current bucket last_modified could be easily retrieved from `getState().bucket.data`.

n1k0: Scratch it, we're speaking deletion here, so we need the last_modified being passed in the action. This comment still actually applies for updates though.

leplatrem:
We could generalize the approach for delete operations too. The only exception would be the deletion from the list of records, which would need to pass the last_modified via the action. Should I go on with delete?

n1k0:
> The only exception would be the deletion from the list of records, which would need to pass the last_modified via the action

Yes, delete actions need to accept an optional last_modified

src/sagas/bucket.js
```diff
@@ -37,10 +37,13 @@ export function* createBucket(getState, action) {
 
 export function* updateBucket(getState, action) {
   const {bid, bucketData} = action;
+  const {bucket: currentBucket} = getState();
+  const {last_modified} = currentBucket;
+  const savedBucket = {...bucketData, last_modified};
```
n1k0: Nit: this gives the false impression the bucket is already saved; `updatedBucket`?

PR #259, opened 2016-10-06
Patch bucket/collection/group attributes (fixes #256)

### LINKED ISSUE #256: Bucket/Group/Collection updates should PATCH instead of PUT
To make sure we don't erase fields that are not defined in the UI schema

leplatrem:
@n1k0 r?


n1k0:
r+

PR #269, opened 2016-10-07
[WIP] Fix #210: Conflict resolution view

### LINKED ISSUE #210: Conflict resolution view
When #203 will be implemented, the server will return a 412 response if someone else changed the object in the mean time.

In some previous versions of the Kinto admin, we had a conflict resolution view that we could reuse in this case :)

leplatrem:
Fixes #210 

- [x] Import old code of admin
- [ ] Import old CSS
- [ ] Typeflow for actions
- [ ] Only resolve data conflicts
- [ ] Plug resolution to current action (eg. `bucket.setData(resolved)` or `collection.updateRecord(resolved)`) with new value for if-match header
- [ ] ...


leplatrem:
After discussing with @n1k0, a possible strategy could be:

- Add a new URL like `#/resolve`
- When a conflict occurs in a saga, dispatch an action with the context info and a `resolve` callback function that receive the resolved data
- Add a reducer that put the infos + the callback in the state
- Redirect to the resolve view
- On cancel resolve, return to the form
- On resolve, excecute the callback with the merged data
- The callback could be the original action dispatch but with the new data

almet:
Should we close this?

PR #270, opened 2016-10-10
Fix #266: clean up states

### LINKED ISSUE #266: Do not merge object attributes with state
Some parts of the code is very messy because we extract some fields from the object attributes into the state (e.g. `uiSchema`, bucket `data`, group `members`, ...). But we don't for records for example. 

See:
* https://github.com/Kinto/kinto-admin/blob/a622b3ab9fd490febdcc0b169ae1f08d68a040c3/src/reducers/bucket.js#L48-L51
* https://github.com/Kinto/kinto-admin/blob/a622b3ab9fd490febdcc0b169ae1f08d68a040c3/src/reducers/group.js#L38-L42
* https://github.com/Kinto/kinto-admin/blob/a622b3ab9fd490febdcc0b169ae1f08d68a040c3/src/components/group/GroupEdit.js#L26

I believe it would simplify greatly if every object was treated equally with regards to its state, storing only:
* `data`
* `permissions`
* `id` *(for convenience maybe, but it's in data)*

It would make the access of state fields for any object consistent everywhere, and the manipulation of data fields for specific widgets would be done at the form level only (while loading initial value and before submitting).

Original #204 
Related #265

leplatrem:
* [x] Refactor
* [x] Green tests
* [x] Human tests :)

### DIFF
--- a/src/components/bucket/BucketEdit.js
+++ b/src/components/bucket/BucketEdit.js
@@ -23,15 +23,10 @@ export default class BucketEdit extends Component {
   render() {
     const {params, session, capabilities, bucket} = this.props;
     const {bid} = params;
-    const {busy} = session;
+    const {busy, data: formData} = bucket;
     if (busy) {
       return <Spinner />;
     }
-    const {data} = bucket;
-    const formData = {
-      id: bid,
-      data
-    };
     return (
       <div>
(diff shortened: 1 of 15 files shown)


leplatrem:
@n1k0 r?

n1k0:
r+

PR #275, opened 2016-10-10
Fix #263: manage objects permissions

### LINKED ISSUE #263: Manage objects permissions
Along the other tabs in objects, we should have a *permissions* panel that allows to see/edit permissions

leplatrem:


css/styles.css
```diff
@@ -341,6 +341,18 @@ table.record-list td.load-more {
   margin: 0;
 }
 
+/* Permissions form table */
```
Natim: Is it commented on purpose?

n1k0:
While working on fixinf a minor styling issue with action buttons, I've added a button to edit record permissions from the list:

leplatrem:
r+

\o/

PR #280, opened 2016-10-13
Ref #263: hide permissions form if readonly

leplatrem:
@n1k0 r?

### DIFF
--- a/src/components/PermissionsForm.js
+++ b/src/components/PermissionsForm.js
@@ -13,6 +13,14 @@ export default class PermissionsForm extends Component {
   }
 
   render() {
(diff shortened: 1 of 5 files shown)


src/components/PermissionsForm.js
```diff
@@ -13,6 +13,14 @@ export default class PermissionsForm extends Component {
   }
 
   render() {
+    if (this.props.readonly) {
+      return (
+        <div className="alert alert-warning">
+          You don't have the required permission to edit the object permissions.
```
n1k0: Nit: *to edit the permissions for this resource*?

PR #288, opened 2016-10-14
Fix #69: bigger submit form buttons

### LINKED ISSUE #69: Improve the styles of forms wrt action button sizes
Spawned from the comments in #65. 

> In the collection create form, «Add» is bigger than «Submit». I was tempted to click «add», which adds a display field instead of the submit button on the left. — idea: replace «Add» by «+» and move submit to the right

leplatrem:
capture d ecran de 2016-10-14 17-56-29

Some friday night thing...

### DIFF
--- a/css/styles.css
+++ b/css/styles.css
@@ -343,6 +343,14 @@ table.record-list td.load-more {
   margin: 0;
 }
 
+/* Generic forms */
+
+form button[type="submit"].btn-primary {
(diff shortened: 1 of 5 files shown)


n1k0:
I'm not entirely sure, but I'm not expert in UX. r+ if you're sure :)

leplatrem:
Yes, I'm sure it's better, although probably not perfect.

Issue #283, opened 2016-10-13
Rename «history» for server urls to something else


leplatrem:

sunakshi96:
Hey 
Can "Account-Details"  work a substitute for history?

leplatrem:
Sorry I didn't provide much details.
I was referring to this https://github.com/Kinto/kinto-admin/blob/773ef8cad8924bd23f2925a830d87948efb0b656/src/actions/history.js
which stores the server URLs that were entered by the user.

Some possible names for these would be: 
* previous server urls
* server url history
* ...

But just `history` is confusing now that we have pages with the history of objects

PR #291, opened 2016-10-17
Update flow version, add typings to UI components.

n1k0:
WiP.

- [x] Add types to collection related components
- [x] Add types to bucket related components
- [x] Add types to group related components
- [x] Add types to record related components
- [x] Add types to other UI components


src/components/AuthForm.js
```diff
@@ -181,11 +188,25 @@ function extendUiSchemaWithHistory(uiSchema, history, clearHistory) {
...
+    uiSchema: Object,
+    formData: Object,
+  };
+
   defaultProps = {
     history: []
   };
 
-  constructor(props) {
+  constructor(props: Object) {
```
n1k0: This would actually want to be `props: Props` and the Prop type being extracted from below, though having these inline is much less verbose...

src/components/AuthForm.js
```diff
@@ -194,7 +215,7 @@ export default class AuthForm extends Component {
     };
   }
 
-  onChange = ({formData}) => {
+  onChange = ({formData}: {formData: Object}) => {
```
n1k0: It would be nice to manage using the `AuthData` union type here, but couldn't have it working :/

src/components/HistoryTable.js
```diff
@@ -6,6 +10,15 @@ import AdminLink from "./AdminLink";
 
 
 class HistoryRow extends Component {
+  props: {
+    bid: string,
+    entry: Object,
```
n1k0: It may be time to create a type for these.

src/components/HistoryTable.js
```diff
@@ -78,6 +92,12 @@ function FilterInfo({location}) {
 }
 
 export default class HistoryTable extends Component {
+  props: {
+    bid: string,
+    location: RouteLocation,
+    history: Object[],
```
n1k0: Ditto

src/components/HomePage.js
```diff
@@ -1,24 +1,27 @@
...
+import type { SessionState } from "../types";
+
 import React, { Component } from "react";
 
 import Spinner from "./Spinner";
 import AuthForm from "./AuthForm";
 
 
-function ServerProps({serverInfo}) {
+function ServerProps({node}: {node: Object}) {
```
n1k0: Note: this is left vaguely typed on purpose as this is a recursive component which deals with differently shaped nodes.

src/components/HomePage.js
```diff
@@ -28,23 +31,29 @@ function ServerProps({serverInfo}) {
...
           </div>
         </div>}
     </div>
   );
 }
 
 export default class HomePage extends Component {
+  props: {
+    session: SessionState,
+    history: string[],
```
n1k0: Refs #283, this name is really really confusing here

src/components/HomePage.js
```diff
@@ -28,23 +31,29 @@ function ServerProps({serverInfo}) {
...
+    clearHistory: Function,
+    setup: Function,
+    navigateToExternalAuth: Function,
```
n1k0: General remark with function types, would be useful to actually type their signatures :D

src/components/JSONRecordForm.js
```diff
@@ -24,7 +25,14 @@ function validate(json, errors) {
 }
 
 export default class JSONRecordForm extends Component {
-  onSubmit = (data) => {
+  props: {
+    disabled: boolean,
+    record: string, // JSON string representation of a record data
```
n1k0: Instead of a comment we could probably rename the prop here, eg. `recordJson`

src/components/Notifications.js
```diff
@@ -70,11 +84,16 @@ export class Notification extends Component {
   }
 }
 
-export default class Notifications extends Component {
+export default class Notifications_ extends Component {
```
n1k0: Note: component names suffixed with an `_` are so because they were conflicting with type names.

src/components/bucket/BucketAttributes.js
```diff
@@ -6,15 +14,24 @@ import BucketTabs from "./BucketTabs";
 
 
 export default class BucketAttributes extends Component {
-  deleteBucket = (bid) => {
+  props: {
+    params: RouteParams,
+    session: SessionState,
+    bucket: BucketState,
+    capabilities: Object,
```
n1k0: It would be nice being able to type the capabilities object

src/components/bucket/BucketForm.js
```diff
@@ -63,7 +70,11 @@ function DeleteForm({bid, onSubmit}) {
         <Form
           schema={deleteSchema}
           validate={validate}
-          onSubmit={({formData}) => onSubmit(formData)}>
+          onSubmit={({formData}) => {
+            if (typeof onSubmit === "function") {
```
n1k0: I had to add this check in a few places in the code, because flow doesn't seem to follow code paths very accurately; here for instance, we can't end up with onSubmit being anything else than a function.

src/components/group/GroupForm.js
```diff
@@ -128,7 +150,7 @@ export default class GroupForm extends Component {
           {` ${creation ? "Create" : "Update"} group`}
         </button>
         {" or "}
-        <Link to="/">Cancel</Link>
+        <AdminLink name="home" params={{}}>Cancel</AdminLink>
```
n1k0: Side note: we could definitely allow the `params` prop to not being passed here.

src/components/record/RecordForm.js
```diff
@@ -9,12 +17,16 @@ import { canCreateRecord, canEditRecord } from "../../permission";
 import { cleanRecord, linkify } from "../../utils";
 
 
-export function extendSchemaWithAttachment(schema, attachment, edit=false) {
-  if (!attachment.enabled) {
+export function extendSchemaWithAttachment(
+  schema: Object,
+  attachmentConfig: ?{enabled: boolean},
```
n1k0: we should type the `required` property too, and maybe destructure these here

src/components/record/RecordForm.js
```diff
@@ -31,8 +43,12 @@ export function extendSchemaWithAttachment(schema, attachment, edit=false) {
   };
 }
 
-export function extendUiSchemaWithAttachment(uiSchema, attachment) {
-  if (!attachment.enabled || !uiSchema.hasOwnProperty("ui:order")) {
+export function extendUiSchemaWithAttachment(
+  uiSchema: Object,
+  attachmentConfig: ?{enabled: boolean}
```
n1k0: Ditto

n1k0:
I'll probably investigate [https://flowtype.org/docs/third-party.html](flow-typed) as well

n1k0:
Okay, most self-review comments addressed, I'm landing this.

PR #294, opened 2016-10-18
Add ability to filter history pages by resource_name

leplatrem:
@n1k0 r?

### DIFF
--- a/src/actions/bucket.js
+++ b/src/actions/bucket.js
@@ -65,7 +65,7 @@ export function listBucketGroupsSuccess(groups: Object[]): Action {
   return {type: BUCKET_GROUPS_SUCCESS, groups};
 }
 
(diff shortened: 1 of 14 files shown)


src/actions/bucket.js
```diff
@@ -65,8 +65,8 @@ export function listBucketGroupsSuccess(groups: Object[]): Action {
   return {type: BUCKET_GROUPS_SUCCESS, groups};
 }
 
-export function listBucketHistory(bid: string, since: Number): Action {
-  return {type: BUCKET_HISTORY_REQUEST, bid, since};
+export function listBucketHistory(bid: string, params: Object = {}): Action {
```
n1k0: Not sure about the `params` name, which one may confuse with the route params; how about `filters`?

Also, this would probably deserve a dedicated type, eg. `HistoryFilters` :)

PR #295, opened 2016-10-24
Fix #278: Groups in permisions

### LINKED ISSUE #278: Specific widgets for groups in permissions form
We could load the list of groups in the current bucket, and use that list to show them in the permissions form.

Instead of having to introduce a group URI in the *Principals* fields, the user could just tick the checkboxes in front of group names like we did for anonymous/authenticated in #277.

leplatrem:
Fix #278 

capture d ecran de 2016-10-21 18-22-45

* [x] PoC
* [x] Generate UI schema from groups
* [x] Find good pattern to load groups on every permission form
* [x] Rework form data to permissions etc.


leplatrem:
Before going further with fixing test for routes saga data loading, @Natim @n1k0 I would take any early feedback please!

src/permission.js
```diff
@@ -110,46 +111,45 @@ export function canEditRecord(session: SessionState, bucket: BucketState, collec
...
-                  };
-                } else {
-                  return perm;
-                }
-              });
-            } else {
-              principalsList = [...principalsList, {principal, permissions: [permissionName]}];
-            }
-            acc.principals = principalsList;
+        if(principal == EVERYONE) {
```
n1k0: Nit: space after `if`

src/permission.js
```diff
@@ -165,13 +165,16 @@ export function permissionsToFormData(permissionsObject: Object): Object {
...
  */
-export function formDataToPermissions(formData: Object) : Object {
-  const {anonymous, authenticated, principals} = formData;
-  const permissionsList = [
-    ...principals,
+export function formDataToPermissions(bid: string, formData: Object) : Object {
+  const {anonymous, authenticated, groups, principals} = formData;
+  const fromGeneric = [
     {principal: EVERYONE, permissions: anonymous},
     {principal: AUTHENTICATED, permissions: authenticated}];
```
n1k0: Nit: please move the closing square bracket to a new line (general coding style)

src/permission.js
```diff
@@ -252,6 +270,14 @@ export function preparePermissionsForm(permissions: string[]) {
         }
       }
     },
+    groups: {
+      classNames: "field-groups",
+      ...groups.reduce((acc, group) => {
+        const {id: gid} = group;
+        acc[gid] = {"ui:widget": "checkboxes"};
```
n1k0: Nit: I believe `acc[group.id] = {"ui:widget": "checkboxes"};` would be okay.

Also, you can return everything immediately:

```js
return {...acc, [group.id]: {"ui:widget": "checkboxes"}};
```

Natim:
feedback + I like the approach

src/sagas/route.js
```diff
@@ -65,11 +67,12 @@ export function* loadRoute(params) {
     });
     // Map them to state
     const bucket     = bid ?               responses[0] : null;
-    const collection = bid && cid ?        responses[1] : null;
-    const group      = bid && gid ?        responses[1] : null;
-    const record     = bid && cid && rid ? responses[2] : null;
+    const groups     = bid ?               responses[1].data : null;
```
n1k0: Could we avoid having this trailing `.data` here?

n1k0:
And re. the way we load route resources, I suggest we file a new separate *tech-debt* issue to properly deal with it.

PR #296, opened 2016-10-25
Fix #258 - Do not try to guess permissions

### LINKED ISSUE #258: Kinto-Admin doesn't detect when people are in groups to give permissions

leplatrem:
If permissions endpoint is available, use it to show/hide buttons.

If not, always show the buttons, and show server error messages.

With https://github.com/Kinto/kinto/issues/874, we could indeed add group support etc. but I'm not confortable with the current situation where the client reimplements the server-side logics of permissions.

This is just a try, not even tested, to see how it could look like


leplatrem:
I think it's ready now to be reviewed/merged

src/sagas/session.js
```diff
@@ -83,7 +83,11 @@ export function* listBuckets(getState, action) {
     // If the Kinto API version allows it, retrieves all permissions
     if ("permissions_endpoint" in serverInfo.capabilities) {
       const {data: permissions} = yield call([client, client.listPermissions]);
-      buckets = handlePermissions(buckets, permissions);
+      buckets = expandBucketsCollections(buckets, permissions);
+      yield put(sessionActions.permissionsListSuccess(permissions));
+    }
+    else {
+      yield put(notificationActions.notifyInfo("Permissions endpoint is not enabled on server."));
```
leplatrem: Add link to server settings docs ?

leplatrem: Explain consequences ?

leplatrem: I can't figure out good wording. And don't know a clean way to put some html (link) in the notification message (from saga, outside react views)

This could land like this until someone gives a hand ;)

PR #297, opened 2016-10-25
Add safety check for list of collections

leplatrem:
@n1k0 r?

Seems to be a bug from server though:
```
$ http :8888/v1/buckets --auth test:test
HTTP/1.1 200 OK
Access-Control-Expose-Headers: Content-Length, Expires, Alert, Retry-After, Last-Modified, Total-Records, ETag, Pragma, Cache-Control, Backoff, Next-Page
Cache-Control: no-cache, no-store
Content-Length: 107
Content-Type: application/json; charset=UTF-8
Date: Tue, 25 Oct 2016 10:01:42 GMT
Etag: "1477306958028"
Last-Modified: Mon, 24 Oct 2016 11:02:38 GMT
Server: waitress
Total-Records: 2

{
    "data": [
        {
            "id": "source", 
            "last_modified": 1477306958028
        }, 
        {
            "id": "destination", 
            "last_modified": 1476717932394
        }
    ]
}

```

```
$ http :8888/v1/buckets/source/collections --auth test:test 
HTTP/1.1 403 Forbidden
Access-Control-Expose-Headers: Content-Length, Expires, Alert, Retry-After, Last-Modified, ETag, Pragma, Cache-Control, Backoff
Content-Length: 95
Content-Type: application/json; charset=UTF-8
Date: Tue, 25 Oct 2016 10:02:10 GMT
Server: waitress

{
    "code": 403, 
    "errno": 121, 
    "error": "Forbidden", 
    "message": "This user cannot access this resource."
}

```

### DIFF
--- a/src/sagas/session.js
+++ b/src/sagas/session.js
@@ -76,7 +76,7 @@ export function* listBuckets(getState, action) {
       }
     });
     let buckets = data.map((bucket, index) => {
-      const collections = responses[index].body.data;
+      const {data: collections=[]} = responses[index].body;
       return {id: bucket.id, collections};
     });
 


leplatrem:
@n1k0 @Natim actually this is not a bug on the server, but the intended behaviour of the work done with this:
* Parent attributes are now readable if child creation is allowed
* https://github.com/Kinto/kinto/issues/803
* https://github.com/Kinto/kinto/pull/828

Attributes of bucket are readable when group creation is allowed, but it does not allow to list collections for example.

Please send feedback in a new kinto issue if you have suggestions...

Natim:
It looks a bit confusing according to me.

leplatrem:
In the tickets I mentionned, we introduced this behaviour: if the user has the permissions to create a child, then we allow to read the attributes of the parent. It was done for collection schema and record creation, but made it consistent for bucket attributes and group/collection creation.

In this case, the user is allowed to create groups. Buckets attributes are readable. Hence the buckets is listed in `/buckets`. But since she is not allowed to create collections, the list of collections is denied.

n1k0:
> since she is not allowed to create collections, the list of collections is denied

I'm surprised the read permission is tied to the create one here, hints?

leplatrem:
>> since she is not allowed to create collections, the list of collections is denied
>>
> I'm surprised the read permission is tied to the create one here, hints?

You can read the parents if you can create children.
Ex. you can read `/buckets` accessible to you if you can create buckets, you can read `/records` accessible to you if you can create records.

Natim:
Ok it makes sense :)

n1k0:
Yeah I get that, on the other hand I was naively thinking that it's not because you can't create children that you can't read them (but I suspect that you can do so if you have the `read` one, right?)

Natim:
Yes you can read them without behing able to create them also.

n1k0:
> since she is not allowed to create collections, the list of collections is denied
> Yes you can read them without being able to create them also.

These two statements are confusing the hell out of me, but I'm just gonna trust you on this one folks 😜

Natim:
In this case, she is not allowed to create collections nor allowed to read collections or any of the existing collections.

Natim:
But maybe because she is able to read the bucket metadata we could return an empty list as we do for `/buckets`?

leplatrem:
> But maybe because she is able to read the bucket metadata we could return an empty list as we do for /buckets?

I don't understand this last one.

Also, this would be worth discussing on the Kinto repo, which possible improvements on documentation, instead of on top a harmless safety check ;)

n1k0:
In any case if we seem confused as kinto team members, we should probably clarify/simplify all this in the docs 😬

leplatrem:
So I close this PR?

Natim:
No I think it doesn't hurt add this change.

PR #300, opened 2016-10-26
Prepare v1.4.1

n1k0:
- Reintroduce `KINTO_ADMIN_PUBLIC_PATH` env var.
- Restore the build command as well as the build/ build target dir.
- Move plugin css files to their own subdir in `css/`.
- Updated README.

r=? @Natim @leplatrem


README.md
```diff
@@ -233,6 +233,14 @@ For instance:
 If you use `target.merkle.tree.file.name` it will render the string
 `foobar` and `target.proof.hash` will render `abcd`.
 
+### How to specify the public root URL for public assets?
+
+If you're not using create-react-app as suggested in the [Installation](#installation) section, you can build a static, production-ready version of the admin using a custom public asset path root by specifying the `KINTO_ADMIN_PUBLIC_PATH` environment variable and running the `npm run build` command:
+
+```
+$ KINTO_ADMIN_PUBLIC_PATH='/admin/' npm run build
```
Natim: Does it means that we will have the signoff plugin activated in that case? What about the max item per page?

n1k0: Ah, nope indeed. So maybe reintroducing this isn't a good idea and using C.R.A should be the only supported way. Thoughts?

PR #306, opened 2016-11-02
Fix #303: Fix attachment link in records list.

### LINKED ISSUE #303: In the list of record clicking the attachment link is invalid.
It is linking to `https://kinto-writer.stage.mozaws.net/v1/admin/fennec/catalog/f0459232-7b26-4193-b568-5212ad6e858a.gz`
when it should be `https://net-mozaws-stage-kinto-fennec.s3.amazonaws.com/fennec/catalog/f0459232-7b26-4193-b568-5212ad6e858a.gz`

It is like if the `base_url` from the home page `attachment.base_url` wasn't used.

n1k0:
r=@Natim?

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -201,7 +201,7 @@ export function humanDate(since: string | number): string {
 }
 
 export function buildAttachmentUrl(record: RecordData, capabilities: Capabilities): ?string {
-  if (!record.attachment || typeof capabilities.attachments !== "object") {
+  if (typeof record.attachment !== "object" || typeof capabilities.attachments !== "object") {
     return;
   }
   const {base_url=""} = capabilities.attachments;
--- a/src/utils.js
+++ b/src/utils.js
@@ -1,6 +1,6 @@
 /* @flow */
 
-import type { RecordData } from "./types";
+import type { RecordData, Capabilities } from "./types";
 import React from "react";
 import _timeago  from "timeago.js";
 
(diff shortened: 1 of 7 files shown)


src/utils.js
```diff
@@ -199,3 +199,12 @@ export function humanDate(since: string | number): string {
...
   }) + " UTC";
 }
+
+export function buildAttachmentUrl(record: RecordData, capabilities: Capabilities): ?string {
+  if (!record.attachment || typeof capabilities.attachments !== "object") {
+    return;
+  }
+  const {base_url} = capabilities.attachments;
+  const {location} = record.attachment;
+  return location.startsWith("http") ? location : base_url + location;
```
Natim: What about `return location.startsWith("base_url") ? location : base_url + location;`?
Can we do something clever so that we make sure that `location` starts with a / and `base_url` doesn't end with one?

test/utils_test.js
```diff
@@ -166,3 +167,30 @@ describe("humanDate", () => {
...
+      .to.be.a("undefined");
+  });
+
+  it("should return nothing if the attachments capability isn't enabled", () => {
+    expect(buildAttachmentUrl({attachment: {}}, {}))
+      .to.be.a("undefined");
+  });
+
+  it("should build an attachment url from the capability config", () => {
+    expect(buildAttachmentUrl({attachment: {location: "/file"}}, {attachments: {base_url: "http://cdn"}}))
```
Natim: I would add some other tests:

```
expect(buildAttachmentUrl({attachment: {location: "file"}}, {attachments: {base_url: "http://cdn"}}))
     .eql("http://cdn/file");
expect(buildAttachmentUrl({attachment: {location: "/file"}}, {attachments: {base_url: "http://cdn/"}}))
     .eql("http://cdn/file");

src/utils.js
n1k0: > What about return location.startsWith("base_url")

You mean testing for the variable, not this string, right?

> Can we do something clever so that we make sure that location starts with a / and base_url doesn't end with one?

I know you don't, but if you ask me, this should be handled by the attachment plugin (also the plugin could always make the `location` field exposing a full url based on the `base_url` config flag methinks).

Anyway, yes I'll add supplementary checks for this.

n1k0:
I've updated the links for attachment previews in https://github.com/Kinto/kinto-admin/pull/306/commits/a863e2de85468ded39b4443bbd218db39eb81513

PR #308, opened 2016-11-03
Fix #307: Always show hooks on collection records page

### LINKED ISSUE #307: Action bar hooks are not shown if no perm to create records
It's understandable to hide `Create` and `Bulk create` when the user has no permission to create records (permission endpoint enabled). 
But hooks should be always shown

leplatrem:
@n1k0 r?


n1k0:
r+

PR #310, opened 2016-11-08
Fix #309: force server URL from build

### LINKED ISSUE #309: Add ability to force server URL when building the admin bundle
**Use case:** be able to send absolute links to particular objects (eg. send an email with the history entries of a particular object).

When we ship the kinto-admin as a server plugin, we could restrict the ability to change the server URL, since we are accessing it via `https:///$SERVER_URL/v1/admin/`.

Alternative approach: provide the server url in the link.

leplatrem:
@n1k0 something like that?

### DIFF
--- a/index.js
+++ b/index.js
@@ -11,6 +11,7 @@ const corePlugins = [
 
 const settings = {
   maxPerPage: 200,
+  singleServer: null,
 };
 
 render(<KintoAdmin plugins={corePlugins} settings={settings} />,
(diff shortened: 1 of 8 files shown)


index.js
```diff
@@ -11,6 +11,8 @@ const corePlugins = [
 
 const settings = {
   maxPerPage: 200,
+  serverURL: "https://kinto.dev.mozaws.net/v1/",
```
Natim: Is there a way to default to the current hostname if we are in `/v1/admin/`?

n1k0: Should be possible by parsing `document.location` indeed.

n1k0:
Looks reasonable :)

leplatrem:
The setting now accepts a string, that we will set with the document.location in the kinto-admin python plugin.
Ready to be reviewed/merged

PR #320, opened 2016-11-16
Fix #318: querystring now takes resource_name into account

### LINKED ISSUE #318: [Signoff] Changes list view doesn't filter records properly
It appears that the HTTP request issued to the history endpoint doen't include the `resource_name` query string parameter: `/v1/buckets/staging/history?collection_id=gfx&_sort=-last_modified&_since=0`

So we end up with all information about the collection as well:

cc @leplatrem

leplatrem:
Fixes #318

### DIFF
--- a/src/routes.js
+++ b/src/routes.js
@@ -74,13 +74,13 @@ function onCollectionRecordsEnter(store: Object, {params}) {
 
 function onCollectionHistoryEnter(store: Object, {params}) {
   const {bid, cid} = params;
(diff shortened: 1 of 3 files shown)


n1k0:
Just a day after thought, maybe we should add a few unit tests for these route hook functions somewhere?

leplatrem:
Should we export them? Otherwise I have no idea how to test that.

n1k0:
Yeah exporting them seems to be the easiest way to go

n1k0:
LGTM r+ with a green build

PR #324, opened 2016-11-17
Fix #316: minor relative time fixes

### LINKED ISSUE #316: [Signoff] Errorenous time representations
Some dates are clearly wrong and/or poorly formatted: 

screenshot at 16 40 11

leplatrem:
I'm not super enthusiastic about the fix I made for relative time in the future. The correct way to go would be to query the server to get a clock diff and then shift timestamps. 
Since our use case consists in showing timestamps in the past, and that we are responsible engineers with NTP, it should not be a problem to keep the current behaviour. I just added a very simple safety to make sure we don't have string like `In 1 minute` when what we expect is `Just now`.

@n1k0 r?

### DIFF
--- a/src/plugins/signoff/components.js
+++ b/src/plugins/signoff/components.js
@@ -121,11 +121,7 @@ function WorkInProgress({label, canEdit, currentStep, step, requestReview, sourc
 
 function WorkInProgressInfos({lastAuthor, lastUpdated}) {
   if (!lastUpdated) {
-    return (
-      <ul>
-        <li>Never updated</li>
-      </ul>
-    );
+    return <ul><li>Never updated</li></ul>;
   }
   return (
     <ul>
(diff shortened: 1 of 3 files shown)


src/plugins/signoff/components.js
```diff
@@ -120,6 +120,13 @@ function WorkInProgress({label, canEdit, currentStep, step, requestReview, sourc
 }
 
 function WorkInProgressInfos({lastAuthor, lastUpdated}) {
+  if (!lastUpdated) {
+    return (
+      <ul>
+        <li>Never updated</li>
+      </ul>
+    );
```
n1k0: Nit: I think you can inline `return <ul><li>Never updated</li></ul>;` here

src/utils.js
```diff
@@ -21,7 +21,12 @@ export function isObject(thing: any): boolean {
 
 export function timeago(date: string | number): string {
   // Show relative time according to current timezone.
-  return _timeago().format(new Date(date));
+  const nowUTC = (new Date()).getTime();
+  // In our use case, we should never show relative time in the future.
+  // For example, if local computer is late, the server timestamp will appear
+  // to be in the future. Hence use "now" as a maximum.
+  const timestamp = nowUTC < date ? nowUTC : date;
```
n1k0: Yeah comparing an int with a string is probably flaky, flow's right!

Issue #326, opened 2016-11-18
Do not show "Connected as X" is user is not returned in hello view


leplatrem:
![screenshot from 2016-11-18 11-40-13](https://cloud.githubusercontent.com/assets/546692/20427352/ce10c51a-ad83-11e6-94c8-e8c8b1fa76ea.png)
![screenshot from 2016-11-18 11-40-54](https://cloud.githubusercontent.com/assets/546692/20427373/e2f0758e-ad83-11e6-83d1-6bc9b4d697df.png)

PR #327, opened 2016-11-18
Fix #313: Configurable auth methods

### LINKED ISSUE #313: Configure the list of available Authentication mecanisms to activate.
In the Kinto Writer we want to Authenticate with LDAP, we should be able to rename Basic Auth in LDAP and hide the FxA authentication method.

leplatrem:
* [x] Basic refactor / implem
* [x] Fix problem ldap submits Authorization as undefined (ref #326)
* [x] Add anonymous method
* [ ] LDAP should return to login form if credentials are invalid (i.e. if user obtained in server info is undefined or does not start with `ldap:`

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -28,10 +28,10 @@ function getAuthHeader(auth: AuthData): ?string {
   }
 }
 
-export function setupClient(session: AuthData): KintoClient {
-  const {server}: {server: string} = session;
+export function setupClient(auth: AuthData): KintoClient {
+  const {server}: {server: string} = auth;
   return setClient(new KintoClient(server, {
-    headers: {Authorization: getAuthHeader(session)}
+    headers: {Authorization: getAuthHeader(auth)}
   }));
 }
 
--- a/src/client.js
+++ b/src/client.js
@@ -13,6 +13,9 @@ function getAuthHeader(auth: AuthData): ?string {
       const {token}: {token: string} = auth.credentials;
       return "Bearer " + token;
     }
+    case "anonymous": {
+      return undefined;
+    }
     // case "ldap": {
     // case "basicauth":
     default: {
--- a/src/client.js
+++ b/src/client.js
@@ -13,7 +13,9 @@ function getAuthHeader(auth: AuthData): ?string {
       const {token}: {token: string} = auth.credentials;
       return "Bearer " + token;
     }
-    case "basicauth": {
+    // case "ldap": {
+    // case "basicauth":
+    default: {
       const {username, password}: {
         username: string,
         password: string,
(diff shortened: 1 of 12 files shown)


leplatrem:
@n1k0 r?

src/reducers/settings.js
```diff
@@ -5,9 +5,10 @@ import type { SettingsState } from "../types";
 const INITIAL_STATE: SettingsState = {
   maxPerPage: 200,
   singleServer: null,
+  authMethods: ["basicauth", "fxa", "ldap"]
 };
 
 // This is basically a noop as we don't support settings update at runtime (yet).
 export default function settings(state: SettingsState = INITIAL_STATE): SettingsState {
-  return state;
+  return {...INITIAL_STATE, ...state};
```
n1k0: Well, that does the same thing no? :) I mean, no action is ever processed here, and so state never changes...

src/types.js
```diff
@@ -280,7 +280,8 @@ export type ServerInfo = {
 
 export type SettingsState = {
   maxPerPage: number,
-  singleServer: ?string
+  singleServer: ?string,
+  authMethods: string[]
```
n1k0: We could create an `AuthMethod` type which would be `"basicauth" | "fxa" | "ldap"`, and here have `AuthMethod[]`

src/client.js
```diff
@@ -13,7 +13,9 @@ function getAuthHeader(auth: AuthData): ?string {
       const {token}: {token: string} = auth.credentials;
       return "Bearer " + token;
     }
-    case "basicauth": {
+    // case "ldap": {
+    // case "basicauth":
```
n1k0: Nit: except if I'm missing something, I'd leave these uncommented

src/components/AuthForm.js
```diff
@@ -218,32 +259,26 @@ export default class AuthForm extends Component {
 
   constructor(props: Object) {
     super(props);
+    const {settings: {authMethods}} = this.props;
+    const defaultAuth = authMethods[0];
```
n1k0: how about being explicit with just `"basicauth"`?

src/reducers/settings.js
leplatrem: I did that to allow to have default settings (ie. not specified in `KintoAdmin` component)

src/components/AuthForm.js
leplatrem: But if `"basicauth"` is not in the list?

leplatrem: yes, since we make it configurable. And for example we could have a stack where only fxa is allowed

PR #328, opened 2016-11-18
Fix #326: Use server info in server infos bar

### LINKED ISSUE #326: Do not show "Connected as X" is user is not returned in hello view

leplatrem:
IMO the current approach was wrong. We should rely as much as possible on what the server returns.

### DIFF
--- a/src/sagas/session.js
+++ b/src/sagas/session.js
@@ -64,7 +64,7 @@ export function* listBuckets(getState, action) {
     const serverInfo = yield call([client, client.fetchServerInfo]);
     // We got a valid response; officially declare current user authenticated
     // XXX: authenticated here means that we have setup the client, not
-    // that then credentials are valid :/
+    // that the credentials are valid :/
     yield put(sessionActions.setAuthenticated());
     // Store this valid server url in the history
     yield put(historyActions.addHistory(getState().session.server));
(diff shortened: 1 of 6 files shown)


src/sagas/session.js
```diff
@@ -63,6 +63,8 @@ export function* listBuckets(getState, action) {
     // Fetch server information
     const serverInfo = yield call([client, client.fetchServerInfo]);
     // We got a valid response; officially declare current user authenticated
+    // XXX: authenticated here means that we have setup the client, not
+    // that then credentials are valid :/
```
n1k0: s/then/the 

I agree that this seems wrong, we should probably mark a user as officially authenticated only when the server has acknowledged the provided creds. Maybe in another issue.

test/components/App_test.js
```diff
@@ -50,8 +55,8 @@ describe("App component", () => {
       expect(infoBar).to.exist;
 
       const content = infoBar.textContent;
-      expect(content).to.contain(session.server);
-      expect(content).to.contain(session.credentials.username);
+      expect(content).to.contain(session.serverInfo.url);
+      expect(content).to.contain(session.serverInfo.user.id);
```
n1k0: Maybe worth having a test for Anonymous user as well, if that's not too painful

PR #335, opened 2016-11-21
Fix broken timeago helper.

n1k0:
We broke it so now it shows "46 years ago" in the records list. Added tests;

### DIFF
--- a/src/components/collection/CollectionRecords.js
+++ b/src/components/collection/CollectionRecords.js
@@ -28,7 +28,7 @@ class Row extends Component {
       return null;
     }
     const date = new Date(lastModified);
-    return Number.isNaN(date.getTime()) ? null : (
+    return date.toJSON() == null ? null : (
       <span title={date.toISOString()}>{timeago(date.getTime())}</span>
     );
   }
--- a/src/components/collection/CollectionRecords.js
+++ b/src/components/collection/CollectionRecords.js
@@ -28,6 +28,7 @@ class Row extends Component {
       return null;
     }
     const date = new Date(lastModified).toISOString();
+    console.log(date);
     return (
       <span title={date}>{timeago(date)}</span>
     );
--- a/src/components/collection/CollectionRecords.js
+++ b/src/components/collection/CollectionRecords.js
@@ -27,9 +27,9 @@ class Row extends Component {
     if (!lastModified) {
       return null;
     }
-    const date = new Date(lastModified).toISOString();
-    return (
-      <span title={date}>{timeago(date)}</span>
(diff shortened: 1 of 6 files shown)


n1k0:
r=? @leplatrem

src/utils.js
```diff
@@ -19,14 +19,14 @@ export function isObject(thing: any): boolean {
   return typeof thing === "object" && thing !== null && !Array.isArray(thing);
 }
 
-export function timeago(date: string | number): string {
+export function timeago(datesrc: string | number, now: ?Date): string {
   // Show relative time according to current timezone.
-  const nowUTC = (new Date()).getTime();
-  const timestamp = parseInt(date, 10);
+  const nowUTC = (now || new Date()).getTime();
```
leplatrem: Maybe add a comment to explicit that it's used for testing?

test/utils_test.js
```diff
@@ -201,3 +202,19 @@ describe("buildAttachmentUrl", () => {
...
 });
+
+describe("timeago", function() {
+  it("should convert a timestamp", () => {
+    expect(timeago(new Date().getTime() - 86400000)).eql("1 day ago");
+  });
+
+  it("should convert a date string", () => {
+    expect(timeago("2016-12-01T00:00:00.000Z",
+           new Date("2016-12-02T00:00:00.000Z"))).eql("1 day ago");
```
leplatrem: I looked in `CollectionRecords` and maybe we don't have to support passing strings. I think we only want to support millisecond timestamps.

n1k0: I discovered we had date strings passed upon inspection (it's in `get lastModified()` where we do `const date = new Date(lastModified).toISOString();`)

n1k0:
I'll definitely remove support for date strings, what a freaking mess...

src/components/collection/CollectionRecords.js
```diff
@@ -27,9 +27,9 @@ class Row extends Component {
     if (!lastModified) {
       return null;
     }
-    const date = new Date(lastModified).toISOString();
-    return (
-      <span title={date}>{timeago(date)}</span>
+    const date = new Date(lastModified);
+    return Number.isNaN(date.getTime()) ? null : (
```
n1k0: `date.toJSON()` seems to return `null` on invalid dates, I'll use that instead.

n1k0:
It seems travis has stopped running our tests on this projectn, because of a pip issue:

Natim:
I know how to fix that, remove the `whoami` in .travis.yml: `pip install --user `whoami` virtualenv`

PR #337, opened 2016-11-21
Fix #336: Fix cannot save bucket attributes upon creation.

### LINKED ISSUE #336: Can't create bucket metadata
1. https://kinto.github.io/kinto-admin/
2. Auth
3. Create bucket
4. Fill some attributes in the JSON editor
5. Submit
6. Attributes weren't saved

Updating works though.

n1k0:
r=? @Natim


n1k0:
Builds are broken for something unrelated to this patch. Landing anyway.

PR #341, opened 2016-11-23
Fix #340: Fix broken saved session restore.

### LINKED ISSUE #340: Sessions are not restored anymore
Trying to reload the admin after a session having successfully been established, we get this in the console:

I'm starting worrying with all these missed regressions, we'll have to provide more component tests, possibly browser tests as well...

n1k0:
WiP, refs #340. Regression has been introduced by #327. It appears we don't have enough typings in `src/index.js` and sagas to ensure full proper checking. 

I'll try to fix this situation.

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -6,7 +6,6 @@
 ./node_modules/
 
 [libs]
-./node_modules/iflow-react-router/index.js.flow
 ./node_modules/iflow-react-router-redux/index.js.flow
 ./interfaces/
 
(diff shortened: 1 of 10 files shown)


n1k0:
r=? @leplatrem

n1k0:
> I cannot see what changed that would fix the regression

```diff
@@ src/index.js
-      this.store.dispatch(sessionActions.setup(session));
+      this.store.dispatch(sessionActions.setup(session.auth));
```

PR #342, opened 2016-11-23
Fix #339: Fix server history not saved.

### LINKED ISSUE #339: Server history dropdown doesn't work anymore
For some strange reason this appears to not work anymore.

/cc @ticabri

n1k0:
r=? @leplatrem

### DIFF
--- a/src/sagas/session.js
+++ b/src/sagas/session.js
@@ -67,7 +67,7 @@ export function* listBuckets(getState, action) {
     // that the credentials are valid :/
     yield put(sessionActions.setAuthenticated());
     // Store this valid server url in the history
-    yield put(historyActions.addHistory(getState().session.serverInfo.url));
+    yield put(historyActions.addHistory(getState().session.auth.server));
     // Notify they're received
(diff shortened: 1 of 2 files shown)


src/sagas/session.js
```diff
@@ -67,7 +67,7 @@ export function* listBuckets(getState, action) {
     // that the credentials are valid :/
     yield put(sessionActions.setAuthenticated());
     // Store this valid server url in the history
-    yield put(historyActions.addHistory(getState().session.serverInfo.url));
+    yield put(historyActions.addHistory(getState().session.auth.server));
```
leplatrem: Here we could save `serverInfo.url` in the history? (i.e. not the URL entered but the actual one defined by the server, e.g. after redirect etc.)
What do you think?

n1k0: This is what was done before but this can be empty when the session has not received this information just yet.

Now I think about it it should never happen, but that's what I saw in the redux logs... I'll investigate

PR #343, opened 2016-11-24
Experiment: Add typings to sagas.

n1k0:
WiP, experimental. I don't much like we're now forced to ensure yielded values are not null; actually I don't understand why is that.

Feedback=? @leplatrem

### DIFF
--- a/interfaces/external-modules/kinto-http.d.js
+++ b/interfaces/external-modules/kinto-http.d.js
@@ -1,15 +1,29 @@
 declare module "kinto-http" {
-  declare type ObjectResponseBody = {
-    data: Object,
-    permissions: Object,
-  }
-
-  declare type ListResponseBody = {
-    data: Object[],
-  }
-
   declare type Options = Object;
-  declare type Permissions = Object;
+  declare type PermissionType =
+      "write"
+    | "read"
+    | "bucket:create"
+    | "group:create"
+    | "collection:create"
+    | "record:create";
+  declare type Principal = string;
+  declare type Permissions = {[key: PermissionType]: Principal[]};
+
+  declare type Resource = {
+    id: string,
+    last_modified: number,
+    [key: string]: any,
+  };
+
+  declare type ObjectResponseBody<T> = {
+    data: T,
+    permissions: Permissions,
+  }
+
+  declare type ListResponseBody<T> = {
+    data: T[],
+  }
 
   declare class KintoClient {
     remote: string;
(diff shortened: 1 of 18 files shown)


interfaces/external-modules/kinto-http.d.js
```diff
@@ -1,21 +1,46 @@
 declare module "kinto-http" {
+  declare type ObjectResponseBody = {
+    data: Object,
+    permissions: Object,
```
leplatrem: Maybe used the `Permission` from below?

interfaces/external-modules/redux-saga.d.js
```diff
@@ -0,0 +1,485 @@
...
+    fork: ForkFn,
+    spawn: SpawnFn,
+    join: JoinFn,
+    cancel: CancelFn,
+    select: SelectFn,
+    actionChannel: ActionChannelFn,
+    race: RaceFn,
+    cancelled: () => CancelledEffect,
+  }
+}
```
leplatrem: This can't be «installed» instead ?

n1k0: This is an import of an existing type definition, so I don't know nor if I really want to start maintaining a diffing version...

src/sagas/bucket.js
```diff
@@ -133,35 +136,41 @@ export function* deleteCollection(getState, action) {
...
-    const {data} = yield call([bucket, bucket.listHistory], {
+    const result = yield call([bucket, bucket.listHistory], {
       since,
       filters: {
         resource_name,
         exclude_resource_name: "record"
       }
     });
-    yield put(listBucketHistorySuccess(data));
+    if (result == null) {
```
n1k0: It looks like a limitation/bug in flow... https://flowtype.org/try/#0PQKgBAAgZgNg9gdzCYAoVUCuA7AxgFwEs5tl8BTAZ3wCYAKASgC4wBxc7cgJwEN84uAHmyYAtgCNuAGhAyAfGADeqMGACehcjAAmYAIwBuFes06wNIwF90uEtTCkAvGArV6DAHScAHvkZHUW2x7HhYRCW4wZ2wPADceGExyIyD7cRZYuEJdAB8HMUkuKIc4hKSDIA

n1k0: So here's why 



I'm afraid we can't do much about it as that's the "expected" behavior in js...

interfaces/external-modules/redux-saga.d.js
leplatrem: You mean, it's not published anywhere? I mean, can we avoid having this huge file in our repo

src/sagas/bucket.js
leplatrem: Hum, that's strange. And why those two methods only then?

n1k0: Because these are the only ones extracting consts from yielded values.

interfaces/external-modules/redux-saga.d.js
n1k0: It's published on flow-typed https://github.com/flowtype/flow-typed but it basically copies the file locally so you can version it, so that's the same

src/sagas/bucket.js
n1k0: Hmm flow v0.36 doesn't raise these errors anymore, so I've removed the checks https://github.com/Kinto/kinto-admin/pull/343/commits/43e1ee0744b12c5e3d6640fbb3eb7d530e0c3f12

it may be due to this bugfix which has landed in 0.34 (we were using 0.33 previously) https://github.com/facebook/flow/issues/2589

I'm stil lfeeling there's too much cargo cult ongoing here, but oh well.

n1k0:
Pair-reviewed with @leplatrem

PR #344, opened 2016-11-28
Add container types.

n1k0:
r=? @Natim @leplatrem

### DIFF
--- a/src/actions/bucket.js
+++ b/src/actions/bucket.js
@@ -12,7 +12,6 @@ import type {
 } from "../types";
 
 import {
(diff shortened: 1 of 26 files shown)


src/actions/bucket.js
```diff
@@ -30,10 +29,6 @@ import {
 } from "../constants";
 
 
-export function bucketBusy(busy: boolean): Action {
-  return {type: BUCKET_BUSY, busy};
-}
-
```
n1k0: Removed this as it was never used.

PR #347, opened 2016-12-02
Improve typing of group state.

n1k0:
WiP. Playing with the awesome `flow coverage` command I've discovered we've been relying on some really generic types here and there.


Natim:
@n1k0 wants to add Records and Collection before landing it.

n1k0:
Won't do.

glasserc:
Why not?

n1k0:
Yeah indeed. Will make a general pass over types.

PR #351, opened 2016-12-07
Fix #208: Paginate history.

### LINKED ISSUE #208: Generalize pagination everywhere
- List of collections in bucket
- List of groups in bucket
- Every history pages

n1k0:
WiP, feedback=? @leplatrem


n1k0:
The patch is ready for review and has been deployed to gh-pages.

r=? @leplatrem

index.js
```diff
@@ -9,7 +9,9 @@ const corePlugins = [
   signoffPlugin
 ];
 
-const settings = {};
+const settings = {
+  maxPerPage: 10,
+};
```
leplatrem: leftover?

src/sagas/bucket.js
```diff
@@ -144,23 +144,39 @@ export function* listBucketCollections(getState: GetStateFn, action: ActionType<
...
 
+export function* listNextHistory(getState: GetStateFn): SagaGen {
+  const {bucket: {listNextHistory}} = getState();
+  if (listNextHistory == null) {
+    return;
+  }
+  try {
+    const {data, hasNextPage, next} = yield call(listNextHistory);
+    yield put(actions.listBucketHistorySuccess(data, hasNextPage, next));
+    yield call([window, window.scrollTo], 0, window.document.body.scrollHeight);
```
leplatrem: note: I think we have that in several places. Wouldn't it better to have an «window action» called like `scrollBottom` and a single place where we actually do this window call?

index.js
n1k0: Yes but that makes me think: if we don't specify a `limit` we're gonna skip the feature and paginate by server defaults, which may be high and inconvenient to use.

Is it okay of should I create a distinct `historyMaxPerPage` setting, so we can have something usable by default in the admin?

src/sagas/bucket.js
n1k0: That's a good idea, will do

index.js
n1k0: Or maybe we can limit every page to 200 items, as we did already in the past?

n1k0: I'm gonna do that and land this.

PR #367, opened 2016-12-30
Introduce a TagsField component.

n1k0:
WiP. This allows entering a comma-separated list of strings in a text input in order to fill a jsonschema array field.

### Usage

For a given json schema declaring an array of strings, set its `ui:field` uiSchema to `tags`:

```js
const schema = {
  "type": "object",
  "properties": {
    "tags": {
      "type": "array",
      "uniqueItems": true,
      "items": {
        "type": "string"
      }
    }
  }
}

const uiSchema = {
  "tags": {
    "ui:field": "tags"
  }
}
```

### Todo

- [x] Initial implementation
- [x] Label support
- [x] Unique values
- [x] Custom separator

### DIFF
--- a/src/components/TagsField.js
+++ b/src/components/TagsField.js
@@ -1,17 +1,37 @@
+/* @flow */
 import React, { Component } from "react";
 
 
+type Props = {
+  schema: Object,
+  uiSchema: Object,
+  formData: string[],
+  onChange: (value: string[]) => void,
+  required: bool,
+  readonly: bool,
+};
+
+type State = {
+  tagsString: string,
+}
+
 export default class TagsField extends Component {
+  props: Props;
+  state: State;
+
   static defaultProps = {
     formData: [],
+    required: false,
+    readonly: false,
   };
 
-  constructor(props) {
+  constructor(props: Props) {
     super(props);
+    console.log(props);
     this.state = {tagsString: props.formData.join(", ")};
   }
 
-  onChange = ({target: {value: tagsString}}) => {
+  onChange = ({target: {value: tagsString}}: {target: {value: string}}) => {
     const tags = tagsString.split(",")
       .map(tag => tag.trim())
       .filter(tag => tag !== "");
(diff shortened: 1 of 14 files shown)


n1k0:
r=? @Natim

src/components/TagsField.js
```diff
@@ -0,0 +1,70 @@
+/* @flow */
+import React, { Component } from "react";
+
+
+function toTagList(tagsString: string, unique: bool = false): string[] {
+  const list = tagsString.split(",")
```
Natim: Should we split on both comma and spaces?

src/components/TagsField.js
```diff
@@ -0,0 +1,70 @@
+/* @flow */
+import React, { Component } from "react";
+
+
+function toTagList(tagsString: string, unique: bool = false): string[] {
```
Natim: Can we add an unittest for this specific `toTagList` fucntion? I have the feeling that we will need to handle specific edge case with time.

Natim:
It looks great :)

n1k0:
@Natim r=?

n1k0:
I've also updated the way tags are rendered in the records list:

test/components/TagsField_test.js
```diff
@@ -0,0 +1,74 @@
...
+      const node = createComponent(TagsField, {
+        schema: {},
+        formData: ["a", "b", "c"],
+      });
+
+      expect(node.querySelector("input").value)
+        .eql("a, b, c");
+    });
+
+    it("should accept and use a custom separator", () => {
```
Natim: Can we add a test with space as a separator?

n1k0: Why? if it works with `:` it works with space as well.

Natim: In that case we are separating with ": " will we separate with two spaces if we use space as a separator?

n1k0: https://github.com/Kinto/kinto-admin/pull/367/commits/314ec4cae005bce2c18acbadcd2b2d710448d49c

Issue #369, opened 2017-01-18
Kinto admin shouldn't fail if /buckets return a 403


Natim:

leplatrem:
Fixed.

PR #370, opened 2017-01-18
Don't fail if buckets return a 403.

Natim:
Refs #369

### DIFF
--- a/test/sagas/session_test.js
+++ b/test/sagas/session_test.js
@@ -178,7 +178,7 @@ describe("session sagas", () => {
         it("should support 403 errors when fetching list of buckets", () => {
           // listBucket fails with unauthorized error.
           const error = new Error();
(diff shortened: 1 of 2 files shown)


test/sagas/session_test.js
```diff
@@ -161,17 +161,44 @@ describe("session sagas", () => {
...
+          listBuckets.next();           // mark user authenticated
+          listBuckets.next();           // add server to history
+          listBuckets.next();           // dispatch server info success
+        });
+
+        it("should support 403 errors when fetching list of buckets", () => {
+          // listBucket fails with unauthorized error.
+          const error = new Error();
+          error.message = "HTTP 403";
+          listBuckets.throw(error);
```
n1k0: Nit: just

```js
listBucket.throw(new Error("HTTP 403"));
```

PR #372, opened 2017-01-20
Fix #248: Prevent crash on uiSchema validation when JSON schema is invalid

### LINKED ISSUE #248: Crash when JSON schema has invalid syntax
Spotted with @n1k0 when submitting a JSON schema with a trailing comma in a list

leplatrem:
Fix #248


leplatrem:
Paired with @n1k0

Issue #378, opened 2017-01-31
Cannot drop a record property


n1k0:
Because we perform PATCH requests, the sent record payload is always merged with its server version, so the property value always remains.

~~We could merge form values into initial record object when sending the request.~~

As patching is only required for collection attributes, so we don't lose attributes not handled by the edit form, we should:

1. Drop our use of the `patch` option for updating buckets, groups and records (so we always replace records);
1. Update the update logic for collections to always preserve existing fields when they initially exist on the server.

Spawned from #377.

n1k0:
Fixed by #379.

PR #379, opened 2017-01-31
Fix #377, #378: Allow dropping edited resource properties.

### LINKED ISSUE #377: Remove property from existing entry in collection
For our A/B experiments we have a collection of entries that look like this (simplified):

```json
{
    "name":"offline-cache",
    "last_modified":1482264657139,
    "id":"5e4277e0-1029-ea14-1b74-5d25d301c5dc",
    "match": {
        "appId":"^org.mozilla.fennec.*$"
    }
}
```

Some experiments we limit to specific app versions using for example the "appId" property in the example above. If we decide to later remove this restriction then we can't do this with the web admin. We could clear it but then we still have this empty property in the JSON and we need to special case this situation in our app. Instead it would be super helpful to be able to remove a property (instead of just clearing it).

Workaround in Fennec:
https://bugzilla.mozilla.org/show_bug.cgi?id=1335355

n1k0:
Refs #377, #378. Deployed to gh-pages.

r? @leplatrem

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "react-breadcrumbs": "^1.3.16",
     "react-codemirror": "^0.2.3",
     "react-dom": "^15.2.0",
-    "react-jsonschema-form": "^0.41.2",
+    "react-jsonschema-form": "^0.42.0",
     "react-redux": "^4.0.0",
     "react-router": "^2.7.0",
     "react-router-redux": "4.0.5",
--- a/src/components/HomePage.js
+++ b/src/components/HomePage.js
@@ -5,6 +5,7 @@ import React, { Component } from "react";
 
 import Spinner from "./Spinner";
 import AuthForm from "./AuthForm";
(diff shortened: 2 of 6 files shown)


n1k0:
> (fields can be lost, and collection review data cannot be removed so it may fail)

I couldn't reproduce the loss of fields, can you point me at a reproducible scenario?

leplatrem:
> can you point me at a reproducible scenario?

Indeed, I made a quick test and it seems to behave properly on collection metadata. How do we manage the difference between attributes that must removed versus. attributes that are missing from data and must remain intact ?

n1k0:
> How do we manage the difference between attributes that must removed versus. attributes that are missing from data and must remain intact ?

It's because initial collection data are passed as *formData* to the form, with preexisting fields kept untouched though submitted, so they're sent as well. Dropping a value by clearing a text field works because we updated to rjsf v0.42.0 which supports this new feature.

PR #380, opened 2017-02-01
Fix #365: Render a JSON diff for history entries.

### LINKED ISSUE #365: Show diff in history entries
Currently the details of history entries for updates are almost useless.

Since each history entry contains the current value, we should instead show a diff (at least a text diff of the stringified JSON) with the preceding history entry.

n1k0:

### DIFF
--- a/src/components/HistoryTable.js
+++ b/src/components/HistoryTable.js
@@ -9,6 +9,7 @@ import { diffJson } from "diff";
 import { timeago, humanDate } from "../utils";
 import AdminLink from "./AdminLink";
 import PaginatedTable from "./PaginatedTable";
+import { getClient } from "../client";
 
 
 function Diff({source, target}) {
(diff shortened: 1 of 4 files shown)


n1k0:
This is wrong, it computes diffs with the immediate previous entry in the list, we need to find the appropriate resource instead. Will rework.

leplatrem:
Yeah, unfortunately because of pagination I guess that the easiest to achieve this is to issue a request when the details panel is opened...

With something like `?uri={uri}&_before={last_modified}&_limit=1`

n1k0:
Would it be hard to eventually have the previous version of the resource attached to the change event on the server?

leplatrem:
> Would it be hard to eventually have the previous version of the resource attached to the change event on the server?

It's more a matter of duplication than complexity... I'd rather be honest, unless someone comes with a nice plan and implements it, we probably won't spend time to change that :|

n1k0:
> I'd rather be honest, unless someone comes with a nice plan and implements it, we probably won't spend time to change that :|

So that's delegating the tedious work to client implementers, that's fine by me but still I feel this would be nice to have through an option or smthg. No big deal, I'll do the atomic request thing to compute the diff.

leplatrem:
> So that's delegating the tedious work to client implementers

Yeah I know... not that the server stuff was not tedious to implement huh ;) 

We discarded the option to duplicate both previous and current in every entry. If we would have done something on the server side, it would have been to store the diffs only. But then the clients would have to apply every diff on top of each other to obtain previous versions. So we found that compromise...

n1k0:
No problem, at least that keeps me busy ;)

src/components/HistoryTable.js
```diff
@@ -71,15 +115,18 @@ class HistoryRow extends Component {
         <tr className="history-row-details"
             style={{display: open ? "table-row" : "none"}}>
           <td colSpan="6">
-            <pre>{JSON.stringify(entry, null, 2)}</pre>
+            {previous
+              ? <Diff source={entry.target} target={previous.target} />
+              : error
+                ? <p className="alert alert-danger">{error}</p>
+                : <p className="alert alert-info">This resource has no previous versions.</p>}
```
leplatrem: When clicking on the details of create, we may want to keep `  <pre>{JSON.stringify(entry, null, 2)}</pre>` no?

src/components/HistoryTable.js
```diff
@@ -4,34 +4,78 @@ import type { ResourceHistoryEntry, RouteLocation } from "../types";
...
-    this.setState({open: !this.state.open});
+    if (this.state.open) {
+      return this.setState({open: false});
+    }
+    if (this.state.previous) {
+      return this.setState({open: true});
+    }
+    fetchPreviousVersion(bid, entry)
+      .then((previous) => this.setState({open: true, previous, error: null}))
+      .catch((error) => this.setState({open: true, previous: null, error}));
```
leplatrem: Maybe we can add a comment here why it is acceptable not to use redux here :)

n1k0:
@leplatrem final r=?

leplatrem:
I really like it :) GG!

PR #391, opened 2017-02-20
Fix #289: Render a loading spinner on bucket list load.

### LINKED ISSUE #289: Show loader in side bar

n1k0:
Refs #289, deployed to gh-pages for testing.

admin-sidebar-loader

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,8 @@
 sudo: false
-language:
-- node_js
-node_js:
-- "6"
+dist: trusty
+language: python
 python:
-- "3.5"
+- "3.6"
 addons:
   apt:
     packages:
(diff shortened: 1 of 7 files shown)


n1k0:
CI is failing because of a python error:



@leplatrem, @natim any idea?

Natim:
Probably because you are running your tests on Kinto master with Python 2 instead of Python 3.

leplatrem:
Nice! 

can you explain the stuff you removed (session busy?) please? (curiosity)

leplatrem:
Ok current tests status is 

```
  447 passing (2s)

  1 failing

  1) collection sagas listNextRecords() Success should scroll to page bottom:

     TypeError: Cannot set property scrollHeight of [object Object] which has only a getter

      at Context.<anonymous> (test/sagas/collection_test.js:144:9)

```

n1k0:
> can you explain the stuff you removed (session busy?) please? (curiosity)

I'm trying to move away from busy state setting trhough dedicated actions, and set the busy flag in the appropriate switch case in the reducer instead, as this has proven more explicit and reliable.

n1k0:
> TypeError: Cannot set property scrollHeight of [object Object] which has only a getter

Wow now I'm really wondering why this happens only now, as we didn't upgrade any dependencies with this patch...

n1k0:
This is especially strange as this doesn't happen locally for me.

PR #393, opened 2017-02-23
Fix #387: Add a full collection changes diff view.

### LINKED ISSUE #387: The review process is currently not really usable
Today you have to look at the history but it really doesn't tells you much.
I would be nicer to see a full diff of what changed.

n1k0:
- [x] full diff view for a collection records list


n1k0:
I'd appreciate early feedback from @leplatrem @natim @glasserc @gabisurita or anybody willing to :)

Natim:
Nice +1

n1k0:
After record deletion:

Natim:
Record deletion is really confusing then, why is it green and not red?

n1k0:
It really depends on the moment you compute the diff from. Here the diff is computed from a starting point where the record which has been deleted didn't exist at all, so there's no way the diff can reflect that. Here's a diff from the very last state before the record is deleted:

Natim:
Ok makes sense.

n1k0:
I've added dedicated navigation buttons to make the history list starting at a change specific timestamp, here's an animation:

tmp

I've deployed the patch to gh-pages so you can play with the feature.

gabisurita:
👍

leplatrem:
Looks nice!

µnit: are there loading spinners while filtering log on a particular entry?

n1k0:
> µnit: are there loading spinners while filtering log on a particular entry?

https://github.com/Kinto/kinto-admin/pull/393/commits/157233f20f9ab6f32e0fdb84f873561e4b3e44ae

n1k0:
final r=? @leplatrem @natim

leplatrem:
👏

Issue #394, opened 2017-03-07
Login doesn't work with Portier


gabisurita:
When I open https://kinto.dev.mozaws.net/v1/admin and switch authentication to Portier, the login flow doesn't work. 

Not sure if this is a Kinto-portier or Kinto-admin issue.

n1k0:
cc @Natim

Natim:
It is probably an issue in kinto-admin because it shows Firefox Account login and then it raises a javascript error.

Natim:
After investigation it is because the kinto-admin doesn't supports Portier auth yet.

n1k0:
Fix ongoing in #395

n1k0:
Fixed by #395

PR #395, opened 2017-03-08
Add support for Portier authentication.

n1k0:
cc @Natim refs #394

### DIFF
--- a/src/actions/session.js
+++ b/src/actions/session.js
@@ -96,7 +96,7 @@ function postToPortier(server: string, redirect: string): NavigationResult {
   // dedicated Kinto server endpoint. This is definitely one of the ugliest
   // part of this project, but it works :)
   try {
-    const portierUrl = (`${server}/portier/login`).replace(/\/\//g, "/");
+    const portierUrl = (`${server}/portier/login`).replace("//portier", "/portier");
     const form: ?HTMLElement = document.querySelector("form.rjsf");
     if (!(form instanceof HTMLFormElement)) {
       return notifyError("Missing authentication form.");
--- a/src/actions/session.js
+++ b/src/actions/session.js
@@ -17,6 +17,8 @@ import {
 } from "../constants";
 
 
+type NavigationResult = ActionType<typeof notifyError> | {type: null};
+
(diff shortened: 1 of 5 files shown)


src/actions/session.js
```diff
@@ -84,17 +84,51 @@ export function logout(): {
...
     const payload = btoa(JSON.stringify(authFormData));
-    const redirect = encodeURIComponent(`${origin}${pathname}#/auth/${payload}/`);
-    document.location.href = `${server}/fxa-oauth/login?redirect=${redirect}`;
+    const redirect = `${origin}${pathname}#/auth/${payload}/`;
+    if (authType === "fxa") {
+      navigateToFxA(server, redirect);
+    } else if (authType === "portier") {
+      postToPortier(server, redirect);
+    } else {
+      return notifyError(`Unsupported auth navigation type "${authType}".`);
```
Natim: 👍

src/components/AuthForm.js
```diff
@@ -188,14 +188,41 @@ const authSchemas = {
...
+        }
+      }
+    },
+    uiSchema: {
+      authType: {
+        ...baseUISchema.authType,
+        "ui:help": (
+          <span>
+            <b>Note:</b> The
+            <a href="https://github.com/Kinto/kinto-portier">{" kinto-portier "}</a>
```
glasserc: What's with this syntax? Isn't `kinto-portier` by itself OK?

n1k0: Just a silly way to ensure spaces wrap the link inner text. React by default don't add spaces between elements separated by text nodes, as in regular HTML. 

So yeah that's ugly, but that prevented me to waste 5 minutes of my life in a CSS stylesheet, which was awesome.

n1k0: If that's confusing, here's a more direct explanation ([jsfiddle](https://jsfiddle.net/n1k0/nfLxq4v6/)):

n1k0: because I'm still feeling bad about that I've asked on twitter about it https://twitter.com/n1k0/status/840256319638654976

Not sure I'll feel better but TGIF.

PR #397, opened 2017-03-09
Fix broken post-auth redirections.

n1k0:
We recently switched from localStorage to sessionStorage for volatile session data persistence, which introduced a regression with redirect URL storage; navigating to external auth sites like fxa and portier clears the sessionStorage, so we were loosing the stored redirectURL. 

This patch reads the redirectURL from the authentication payload we receive as an URL parameters and fixes this issue.

### DIFF
--- a/src/sagas/route.js
+++ b/src/sagas/route.js
@@ -98,25 +98,34 @@ export function* loadRoute(params: RouteParams): SagaGen {
 export function* routeUpdated(getState: GetStateFn, action: ActionType<typeof actions.routeUpdated>): SagaGen {
   const {session: {authenticated}} = getState();
   const {params, location} = action;
-  const {token} = params;
+  const {token, payload} = params;
 
   // Clear notifications on each route update
   yield put(clearNotifications());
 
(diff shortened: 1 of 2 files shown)


src/sagas/route.js
```diff
@@ -98,25 +98,34 @@ export function* loadRoute(params: RouteParams): SagaGen {
...
-    yield put(storeRedirectURL(null));
+  if (!authenticated && location.pathname !== "/") {
+    if (token && payload) {
+      // We've just been redirected with an auth token and payload
+      const {redirectURL} = JSON.parse(atob(payload));
+      // Wait until we're actually authenticated
+      yield take(SESSION_AUTHENTICATED);
+      // Redirect to the initially requested URL
+      yield put(updatePath(redirectURL));
+    } else {
```
leplatrem: nit: as a reader, I wish there was a small comment here to remind me what's going on here in this `else`

n1k0: Couldn't spot how. Please open a PR if you have ideas :)

PR #398, opened 2017-03-13
Prettierify code.

n1k0:
We've been migrating most of our Kinto-related frontend codebase to the [prettier](https://github.com/jlongster/prettier) coding style, and this patch does the same for the admin.


glasserc:
It's cool that Prettier supports flowjs.

PR #400, opened 2017-03-14
Fix #399: Retrieve all paginated permissions.

### LINKED ISSUE #399: Incorrect permissions displayed (BUG?)
The admin does not load the proper permissions for buckets and collections - displaying readonly for many buckets and collections. Tests using the API reveals that the account in question does has write access to these buckets and collections.

It appears this happens because the admin only retrieves the first page of results from `/permissions`, which is may be missing data about the root bucket or collection.

I'm using v1.10.0, but was able to recreate the issue in v1.11.1.

Unfortunately I was not able to track down any bug in the code, but I'm wondering if this is a known issue.

n1k0:
Fixes #399.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     "diff": "^3.2.0",
     "express": "^4.14.0",
     "filesize": "^3.3.0",
-    "kinto-http": "^3.1.0",
+    "kinto-http": "^4.1.0",
     "react": "^15.2.0",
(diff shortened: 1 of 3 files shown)


n1k0:
Deployed patch to gh-pages for easier QA.

n1k0:
Released in [v1.11.2](https://github.com/Kinto/kinto-admin/releases/tag/v1.11.2).

Issue #404, opened 2017-03-15
Links on partial diff view for resource != record? 


leplatrem:
I have this history:
![screenshot from 2017-03-15 16-56-01](https://cloud.githubusercontent.com/assets/546692/23957699/4fd0dd42-09a0-11e7-9c43-086daf73c52c.png)

If I click on every button above the most recent **record entry**, then the resulting view will be empty, which is confusing.

There is something to clarify here: should the partial history include entries of other resources like the actions on collection ?
If not, then the links to partial history should only appear on entry where resource name is record.
If yes, then what would the full diff mean in this context ?

n1k0:
Yeah, we should only provide links next to record related entries, as full diffs for other kind of resources is not supported.

n1k0:
I also realize that we currently only provide full collection diff for *records*, not collection *attributes*. We may want to compute a full diff including them as well, but that would require the ability to comute a snapshot for a single collection object, which kinto-http doesn't provide atm.

n1k0:
Released in [v1.12.0](https://github.com/Kinto/kinto-admin/releases/tag/v1.12.0).

PR #409, opened 2017-03-16
Fix #405: Collection full diff view rewording.

### LINKED ISSUE #405: Show last_modified in full diff ?
In diff for single history entries, we show `last_modified` in the diff.  But the **full** diff we don't.

Maybe *full diff* is not the right term for the feature? I mean, if it contains less info :)

n1k0:
Fixes both #404 and #405.

### DIFF
--- a/src/components/HistoryTable.js
+++ b/src/components/HistoryTable.js
@@ -148,7 +148,8 @@ class HistoryRow extends Component {
           </td>
           <td>{user_id}</td>
           <td className="text-center">
-            {enableDiffOverview &&
+            {resource_name === "record" &&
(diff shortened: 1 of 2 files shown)


src/components/HistoryTable.js
```diff
@@ -148,7 +148,8 @@ class HistoryRow extends Component {
           </td>
           <td>{user_id}</td>
           <td className="text-center">
-            {fullDiffSupport &&
+            {resource_name === "record" &&
```
n1k0: This specifically addresses #404.

n1k0:

PR #410, opened 2017-03-16
Fix #408: Limit number of collections listed in the sidebar.

### LINKED ISSUE #408: Limit the number of bucket collections to render in the sidebar
If a bucket contains hundreds of collections, the UI is unlikely to scale. We should limit the number of collections to render (eg. 5 most recently updated ones), with a link to see all remaining bucket collections.

Note that the same should apply for the list of buckets.

n1k0:
Refs #408

### DIFF
--- a/src/sagas/session.js
+++ b/src/sagas/session.js
@@ -49,7 +49,8 @@ export function* sessionLogout(
 
 export function expandBucketsCollections(
   buckets: BucketEntry[],
-  permissions: PermissionEntry[]
+  permissions: PermissionEntry[],
+  sidebarMaxListedCollections: number
 ): BucketEntry[] {
   // Create a copy to avoid mutating the source object
   const bucketsCopy = clone(buckets);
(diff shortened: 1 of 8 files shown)


n1k0:
Patch deployed to gh-pages.

src/sagas/session.js
```diff
@@ -136,7 +151,9 @@ export function* listBuckets(
     }
     const responses = yield call([client, client.batch], batch => {
       for (const { id } of data) {
-        batch.bucket(id).listCollections();
+        batch
+          .bucket(id)
+          .listCollections({ limit: sidebarMaxListedCollections + 1 });
```
Natim: 👍

n1k0: Yeah wasn't sure... but that allows us to always have one more item than what's in `sidebarMaxListedCollections`, so in the sidebar we can check if the number of items is strictly greater than `sidebarMaxListedCollections` to know if there are more entries on the server, and so we can render a link to paginated list of collections. That's admittedly convoluted, but works great :)

n1k0:
> Maybe we can add a test to check that kinto-http.js is called with the limit

Sadly, we never found a good (non-hellish) way to test for things happening within a `batch` closure...

I'm gonna land the patch.

n1k0:

PR #413, opened 2017-03-20
Fix #407: Order history entry target permissions consistently.

### LINKED ISSUE #407: Order permissions before diff?
We should not show a diff for ther principals who weren't changed:

screenshot from 2017-03-15 17-09-48

I'm not sure that the server guarantees the order of principals in the permissions. Maybe it should (*please open issue*).

In the meantime we could order them alphabetically...

n1k0:
Refs #407, ensure sorting history entry permissions list before rendering a diff.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -1,6 +1,6 @@
 /* @flow */
 
-import type { RecordData, Capabilities } from "./types";
+import type { RecordData, ResourceHistoryEntry, Capabilities } from "./types";
 import React from "react";
 import _timeago from "timeago.js";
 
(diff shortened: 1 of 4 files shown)


n1k0:
Hold on, I'm gonna add a few unit tests for that function introduced in `utils`.

src/utils.js
```diff
@@ -239,3 +239,22 @@ export function scrollToBottom(): Promise<void> {
...
   return Promise.resolve();
 }
+
+export function sortHistoryEntryPermissions(
+  entry: ResourceHistoryEntry
+): ResourceHistoryEntry {
+  if (entry.action === "delete") {
+    return entry;
+  }
+  const { permissions } = entry.target;
```
glasserc: Rather than checking explicitly for `delete`, should we just check for the absence of `permissions`?

n1k0: Good idea, will update.

n1k0:

PR #414, opened 2017-03-20
sagas/route: redirectURL can sometimes be null

glasserc:
redirectURL starts out as null and is only updated if a user happens to
be viewing an authenticated page when they aren't logged in.

If a user tries to log in from the home page, no redirectURL will be
present and this updatePath call will fail.

### DIFF
--- a/src/sagas/route.js
+++ b/src/sagas/route.js
@@ -133,8 +133,10 @@ export function* routeUpdated(
       const { redirectURL } = JSON.parse(atob(payload));
       // Wait until we're actually authenticated
       yield take(SESSION_AUTHENTICATED);
-      // Redirect to the initially requested URL
-      yield put(updatePath(redirectURL));
+      // Redirect to the initially requested URL (if any)
+      if (redirectURL) {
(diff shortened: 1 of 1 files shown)


src/sagas/route.js
```diff
@@ -133,8 +133,10 @@ export function* routeUpdated(
       const { redirectURL } = JSON.parse(atob(payload));
       // Wait until we're actually authenticated
       yield take(SESSION_AUTHENTICATED);
-      // Redirect to the initially requested URL
-      yield put(updatePath(redirectURL));
+      // Redirect to the initially requested URL (if any)
+      if (redirectURL) {
```
n1k0: If we do not proceed with a redirect (at all), this might mean we'll keep the auth token in the URL hash. How about falling back to `"/"` as a default redirect?

n1k0:
Landed as https://github.com/Kinto/kinto-admin/commit/de9d2beaedd0497111abba062eab7f4f8b78ef77, thanks!

n1k0:

Issue #421, opened 2017-04-01
does not appear to work with kinto 7.0.0.dev0


vladp:
new user, so could be my fault, 
but I am getting
```

Couldn't list buckets.
requests.0 in body: Unrecognized keys in mapping: "{'retry': 1, 'batch': True, 'safe': False, 'bucket': 'vsp_firstbucket_v1'}"
HTTP 400 Invalid parameters: Invalid request parameter (requests.0 in body: Unrecognized keys in mapping: "{'retry': 1, 'batch': True, 'safe': False, 'bucket': 'vsp_firstbucket_v1'}")
```

after creating a first bucket. bucket name is vsp_firstbucket_v1

Natim:
ping @gabisurita is this a problem with the swagger validation?

gabisurita:
I can't see how this relates with the swagger validation. Do you have the server logs?

gabisurita:
I investigated this a bit and this is what I've got:

When I try using the admin with a local instance there's no log on the server, which means kinto-admin is probably trying to fetch something else and failing. I'm not familiar with the kinto-admin codebase, but it seems like kinto-admin is somehow trying to enforce SSL and crashing if it fails. 

When I tried using a heroku deployed master with HTTPS it worked well, when I changed the settings to HTTP it also crashed.

Natim:
Thanks you for investigating.

n1k0:
From the initial bug report description it seems like the client is sending the request options as part of the request body, which should never happen and is definitely a bug.

@vladp do you have sample code so we can inspect client calls?

gabisurita:
It came to me that maybe what I investigated and the initial report are two unrelated bugs. Haha

vladp:
@n1k0  the sources is exactly what I had pull from this GitHub repo on saturday, I did not modify a single line.

glasserc:
I was able to reproduce this locally with kinto-admin master and kinto master. I'll investigate.

glasserc:
Some investigation revealed that:

- The reason it's a POST is because the kinto-admin is doing a batch request to request the collections for all extant buckets.
- kinto-http `Bucket#listCollections` calls a method called `_bucketOptions` which copies any options from the client into the request, along with `bucket: "bucketname"` and `batch: false`.
[...]

PR #422, opened 2017-04-10
Bump kinto-http

glasserc:
Pull in the fix to Kinto/kinto-http.js#180 so that we can address #421.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,4 +1,5 @@
 /* @flow */
+import "babel-polyfill";
 import type { Plugin } from "./types";
 
 import React, { Component } from "react";
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     "diff": "^3.2.0",
     "express": "^4.14.0",
     "filesize": "^3.3.0",
-    "kinto-http": "^4.1.0",
+    "kinto-http": "^4.3.0",
     "react": "^15.4.2",
     "react-breadcrumbs": "^1.3.16",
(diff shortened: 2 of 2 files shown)


glasserc:
@n1k0: Do you know if this is the correct way to use a library that is async/await?

glasserc:
And also, can you tell me why the build is failing, even though `babel-polyfill` is part of the `devDependencies` 😆

src/index.js
```diff
@@ -1,4 +1,5 @@
 /* @flow */
+import "babel-polyfill";
```
Natim: I wish we could handle this without having to add babel-polyfill back in the mix.

n1k0: I don't clearly get why we suddenly need this, please refresh my memory

Natim: I guess it is an attempt to fix the tests, also it doesn't so maybe you know what is missing.

n1k0: fwiw `async/await` is supported through some Babel directibe added to `.babelrc` directly, no need for dedicated imports. At least that's what worked in kinto-http.js.

n1k0:
Superseded by #423.

src/index.js
glasserc: I'm not an expert, but it seems like the kinto-http.js artifacts did not include async/await code, but were already transformed into something that relied on a `regeneratorRuntime`. This `regeneratorRuntime` was provided by `babel-polyfill` or the "regenerator runtime", whatever that was.

PR #429, opened 2017-04-26
Fix #428: Fix signoff plugin membership checks.

### LINKED ISSUE #428: [1.13.2] Error when navigating on a collection with the signoff plugin.
I was able to reproduce using the kinto-admin in gh-pages.

error

n1k0:
Deployed to gh-pages. Note: needs enabling the Moz VPN and checks the `https://kinto-writer.stage.mozaws.net/v1/` server.

### DIFF
--- a/src/plugins/signoff/components.js
+++ b/src/plugins/signoff/components.js
@@ -16,7 +16,7 @@ import { ProgressBar, ProgressStep } from "./ProgressBar.js";
 
 function isMember(groupKey, source, sessionState, bucketState) {
   const { serverInfo: { user = {}, capabilities } } = sessionState;
-  if (!user.id) {
+  if (!source || !user.id) {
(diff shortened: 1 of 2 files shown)


n1k0:
Side note: this was a quick fix, would need more investigation to learn how and when this regression happened.

n1k0:
Hold on, new pb: `bundle.js:62 Uncaught TypeError: Cannot read property 'lastAuthor' of undefined`

n1k0:
The right fix is in https://github.com/Kinto/kinto-admin/pull/429/commits/41bf3cd60310278095b44a74542dee44e97d219e

Payload key names were wrong, and we don't have proper types for sagas, which is probably something to fix later, but not here as it's a huge amount of work and we need a release asap.

PR #431, opened 2017-04-27
Update kinto-http.js

Natim:
I don't know if this change is necessary.


leplatrem:
But a new build of the admin plugin in Kinto to include the new version is necessary though

Issue #419, opened 2017-03-24
Pre-commit hook ignores partial commits


leplatrem:
* `git add -p`
* `git commit -m Partial`
* `git status` → everything was committed instead of just the added changes during `git add -p`

Natim:
Outch 🐞

glasserc:
I don't think we will be able to fix this until https://github.com/okonet/lint-staged/issues/62 is resolved.

glasserc:
Discussion in the linked issue is mostly stalled IMHO, with some useful information supplied by the author of http://pre-commit.com/ which actually handles partially-staged files correctly. I've tried that project and liked it, but it requires installing the pre-commit tool, for example using pip, and explicitly running `pre-commit install` when you clone the repo. How would you feel about moving to that? I'll whip up a PR for discussion.

glasserc:
This should be closed by #445.

leplatrem:
Awesome thanks!

PR #445, opened 2017-07-10
Introduce pre-commit to the codebase

glasserc:
Experiment to introduce pre-commit to try to address #419.

To use:

- ``pip install --user pre-commit``
- ``pre-commit install`
- ``rm .git/hooks/pre-commit.legacy``
- Edit some file -- try tweaking spacing to trigger prettier, or introducing an unused variable to annoy eslint
- Stage the file (or only stage part of it -- only staged parts will be linted)
- Try to commit, or ``pre-commit run``, or ``pre-commit run eslint``/``pre-commit run prettier``

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,11 @@ Kinto-based systems.
 
 NodeJS v4+ and npm 2.14+ should be installed and available on your machine.
 
+Various pre-commit hooks are available to help prevent you from
+pushing sub-optimal code; to use these, ``pip install --user
+pre-commit`` and ``pre-commit install``. (If you have a
+``.git/hooks/pre-commit.legacy``, you can remove it.)
+
 ## Installation
 
 The easiest way to get started is to install [create-react-app](https://github.com/facebookincubator/create-react-app) first:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,12 @@
+- repo: local
+  hooks:
+    - id: eslint
+      name: eslint
+      language: system
+      entry: node_modules/.bin/eslint
+      types: [javascript]
+    - id: prettier
+      name: prettier
+      language: system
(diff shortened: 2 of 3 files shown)


glasserc:
Although pre-commit allows/encourages the use of upstream repos as lint "hooks", we don't take advantage of that in this PR, because eslint doesn't expose the right metadata, and prettier hasn't exposed it yet (although will in the next release). If those two things change, we can stop npm installing these tools manually using packages.json and start installing them automatically using pre-commit. That approach would have some trade-offs; we'd have to update our CI infra to run the commands the same way, and greenkeeper would no longer bug us about new versions of prettier and eslint. For now I'm changing as little as possible for the sake of this PR.

README.md
```diff
@@ -31,6 +31,11 @@ Kinto-based systems.
 
 NodeJS v4+ and npm 2.14+ should be installed and available on your machine.
 
+Various pre-commit hooks are available to help prevent you from
+pushing sub-optimal code; to use these, ``pip install --user
```
n1k0: Hmm requiring python seems slightly bold here. No equivalent npm package available?

glasserc: Well, the tool is written in Python, so one way or another, you'll have to install Python. It seems that pre-commit can be installed itself using `brew`, but not everyone uses OS X, and I guess there's a `curl` installation option as well. I didn't worry about it too much because this is a tool for administering Kinto, which is itself written in Python, so I think there's a good chance you'll have Python and even `pip` available on your machine.

glasserc: N.B. There are a couple npm packages called `pre-commit` but they appear to be unrelated to the tool by Yelp.

glasserc:
N.B. without installing the `pre-commit` tool, you can still commit as normal, just without pre-commit hooks. This isn't really an "attack vector" because you can always `git commit -n` to disable existing hooks.

README.md
n1k0: Sure high are chances people have a working python environment. Ship it!

PR #440, opened 2017-06-23
Update some, but not all, of the dependencies

glasserc:
This is meant to get us closer to #433, although it only updates the easy ones. React-Router and friends are still TBD. Hopefully landing this will facilitate that.

PR #455, opened 2017-10-21
Reactivate and fix flow typechecking

glasserc:
This broke in #440, apparently because of https://github.com/facebook/flow/issues/3528, because Sidebar.js uses a spread in an array without a type declaration, which as of 0.39 causes pathological behavior. The form of the breakage is that a child process of the flow typechecker starts to consume an infinite amount of memory, and when that child process dies, the parent reports an error and exits with status code zero indicating success. Thus, travis believes the build has succeeded.

Apparently the way to diagnose this sort of thing is to start marking files as ignored in .flowconfig, rerun flow and see if it terminates within a few seconds (and without consuming an unreasonable amount of memory, here defined as >10% of a Thinkpad X1 Carbon's RAM). This is basically the approach I took in this PR. We successively unignore files until we're back to where we started.

Fixing this uncovered a rat's nest of typechecking that had also been broken. In particular, some changes to the types for React meant that we had to change `Component { props: Foo }` to `Component<Foo>`. There were also a handful of other errors (some spurious, some not) that apparently a modern flow-check can catch. Fix these too.

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,9 +1,8 @@
 [ignore]
 .*/node_modules/*
 .*/test/.*
-.*/src/components/.*
-.*/src/containers/.*
-.*/src/plugins/.*
+.*/src/components/Sidebar.js
+.*/src/containers/Sidebar.js
 .*/src/routes.js
 .*/src/index.js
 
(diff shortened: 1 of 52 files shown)


.flowconfig
```diff
@@ -14,4 +14,3 @@ esproposal.class_static_fields=enable
 unsafe.enable_getters_and_setters=true
 module.name_mapper='.*\(.css\)' -> 'CSSModule'
 module.system=haste
-strip_root=true
```
glasserc: N.B. This is no longer a valid option in .flowconfig, although it continues to be supported as a command-line argument if you need it. It doesn't really change the output much.

Issue #446, opened 2017-07-25
Problems installing kinto-admin


Grahack:
Hi,  
I encountered several problems using the first described process:

* complaints about not finding `babel-polyfill`, after adding it to `package.json` it was OK,
* `ReferenceError: regeneratorRuntime is not defined` at line 24

     21 |  * @return {Promise}
     22 |  */
     23 | var pMap = exports.pMap = function () {
     24 |   var _ref = _asyncToGenerator(regeneratorRuntime.mark(function _callee2(list, fn) {
     25 |     var results;
     26 |     return regeneratorRuntime.wrap(function _callee2$(_context2) {
     27 |       while (1) {

I had no problem with the second described process: `git clone…`.

Thanks for kinto!

glasserc:
Hi, sorry for not writing back sooner! What you describe sounds like a missing babel-generator-runtime. We've had some questions in our minds about how to best package kinto libraries (see Kinto/kinto.js#649 and Kinto/kinto.js#650), but I think kinto-admin (as an executable) should be ready to run "out-of-the-box", which means including babel-polyfill as a dependency rather than just as a devDependency. I've opened #456 to try to address this. However, I'm not sure why you continued to encounter the issue even after adding babel-polyfill. Could you elaborate on where you added it -- your own package.json, or somehow to the kinto-admin one?

leplatrem:
Should fixed on master. Feel free to reopen if problems still occur

PR #456, opened 2017-11-02
Add babel-polyfill to dependencies

glasserc:
Refs #446.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
   },
   "dependencies": {
     "atob": "^2.0.3",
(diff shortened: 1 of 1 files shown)


leplatrem:
@glasserc what's the status of this? Is there any risk to burst out the final bundle size?

glasserc:
I think this PR is correct and should probably be merged, and we should probably close the linked bug because of inactivity, but we should be ready to revert if someone explains that they are using kinto-admin as a component or something strange like that :)

PR #462, opened 2018-02-22
Fix redux-saga deprecated imports

leplatrem:
I first wanted to fix every deprecation warnings, but it seems we are running behind in terms of dependencies, and `git grep` does not give anything relevant :) 

```
Warning: Accessing PropTypes via the main React package is deprecated, and will be removed in  React v16.0. Use the latest available v15.* prop-types package from npm instead. For info on usage, compatibility, migration and more, see https://fb.me/prop-types-docs

Warning: Accessing createClass via the main React package is deprecated, and will be removed in React v16.0. Use a plain JavaScript class instead. If you're not yet ready to migrate, create-react-class v15.* is available on npm as a temporary, drop-in replacement. For more info see https://fb.me/react-create-class
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,7 @@ install:
   - export DISPLAY=':99.0'
   - Xvfb :99 -screen 0 1024x768x24 > /dev/null 2>&1 &
   - npm install
+  - npm ls
   - npm run build
 env:
   - ACTION="build:lib"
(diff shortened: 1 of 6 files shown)


leplatrem:
Ok let's wait for redux-saga 1.0 to be released. Grrr.

PR #466, opened 2018-02-26
Fixes #461: [signoff] Support per-bucket configuration

### LINKED ISSUE #461: [signoff] Support of per-bucket configuration in review
See https://github.com/Kinto/kinto-signer/issues/144

leplatrem:
Ref #461

* [x] Basic idea
* [x] Add tests for placeholders in groups
* [x] Make sure it's sound with server part (see https://github.com/Kinto/kinto-signer/issues/144)

### DIFF
--- a/src/plugins/signoff/sagas.js
+++ b/src/plugins/signoff/sagas.js
@@ -8,15 +8,14 @@ import * as SignoffActions from "./actions";
 export function* onCollectionRecordsRequest(getState, action) {
   const { bid, cid } = action;
   const { session: { serverInfo } } = getState();
-  // See if current collection is among signed resources from capabilities.
-  const { capabilities: { signer = { resources: [] } } } = serverInfo;
-  const resource = signer.resources.filter(r => {
-    return r.source.bucket == bid && r.source.collection == cid;
-  })[0];
+  // See if currently viewed collection is among kinto-signer resources
+  // described in server info capabilities.
+  const resource = _pickSignoffResource(serverInfo, bid, cid);
 
   // Refresh the signoff toolbar (either empty or basic infos about current)
   yield put(SignoffActions.workflowInfo(resource));
 
+  // Current collection is not configured, no need to proceed.
   if (!resource) {
     return;
   }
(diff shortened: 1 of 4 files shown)


leplatrem:
Works like a charm with with https://github.com/Kinto/kinto-signer/pull/215

The plan is:
1. deploy this fist since it's fully backward compatible
2. land and deploy the new kinto-signer with the same configuration
3. change the configuration to have per-bucket review

src/plugins/signoff/sagas.js
```diff
@@ -198,3 +197,24 @@ function _updateCollectionAttributes(getState, data) {
...
+    }
+  )[0];
+  // The whole UI expects to have collection information for source/preview/destination.
+  // If configured by bucket, fill-up the missing attribute as if it would be configured
+  // explicitly for this collection.
+  if (resource && !resource.source.collection) {
+    resource.source.collection = resource.destination.collection = cid;
+    if (resource.preview) {
+      resource.preview.collection = cid;
+    }
```
glasserc: Same situation where you need a deepcopy here.

test/plugins/signoff/sagas_test.js
```diff
@@ -5,177 +5,278 @@ import { notifySuccess } from "../../../src/actions/notifications";
...
+                        bucket: "stage",
+                        collection: "plugins",
+                      },
+                      preview: {
+                        bucket: "preview",
+                        collection: "plugins",
+                      },
+                      destination: {
+                        bucket: "prod",
+                        collection: "plugins",
```
glasserc: Maybe change these collections so we're sure it isn't matching the second case?

Are we actually going to support these sorts of configurations, with data potentially matching multiple signer configs?

leplatrem: > Are we actually going to support these sorts of configurations, with data potentially matching multiple signer configs?

It will take the first one anyway. We can add additional safety checks on the server. https://github.com/Kinto/kinto-signer/issues/216

PR #477, opened 2018-03-05
Update to 4.x series of react-codemirror2

glasserc:


glasserc:
We don't do anything nearly as complicated in this project as we do in the rjsf playground, but let's still try to maintain the same behavior we had with react-codemirror2 v2.

I tested this by hand. #473 didn't fail any tests (although it should have??) because there aren't any tests for the JSONEditor component (except maybe the browser tests?).

leplatrem:
>  #473 didn't fail any tests (although it should have??

Yeah, I wonder how we could unit test that.... browser could have been the key indeed.

PR #482, opened 2018-03-12
Get the list of auth methods supported by the server when first rendering the auth form

### LINKED ISSUE #331: Fetch authentication methods from server capabilities
* User picks a server
* Capabilities are fetched
* Known authenticated methods are filtered against the methods that the server supports

magopian:
Fixes #331


magopian:
This isn't ready yet, it also needs the capabilities to be parsed on each server change (the current `onChange` implementation is just some placeholder for now).

src/components/AuthForm.js
```diff
@@ -363,18 +367,67 @@ export default class AuthForm extends PureComponent<
...
+    const knownAuthImplementations = [
+      "basicauth",
+      "account",
+      "fxa",
+      "ldap",
+      "portier",
+      "anonymous",
+    ];
+    const server = (history && history[0]) || defaultKintoServer;
+    const defaultAuth = "anonymous";
```
leplatrem: nit: this could be a global const

src/components/AuthForm.js
```diff
@@ -363,18 +367,67 @@ export default class AuthForm extends PureComponent<
...
+      // Server changed, request its capabilities and check what auth methods
+      // it supports.
+      setupClient({ authType: "anonymous", server: server });
+      const client = getClient();
+      client
+        .fetchServerInfo()
+        .then(info => {
+          console.log("fetchServerInfo", info);
+        })
+        .catch(() => console.log("error"));
```
leplatrem: nit: use `await` (and possibly reuse code with `componentDidMount`

magopian: yeah, as explained above, this code is only a placeholder (a previous version of the code i didn't mean to commit...) I'll finalize this tomorrow ;)

magopian:
Regarding the use of redux/sagas, i'm not sure... it seems that most of the state is handled in the component when it is **local** (like the auth methods in the `AuthForm` I guess?).

In a project using redux/sagas for everything I would have done it, in this specific case, I'm not sure... What do you think? Also maybe @glasserc has some opinion on that?


Oh, also this PR needs tests.

glasserc:
I'm probably the wrong person to ask about these things, because I don't really understand why components in React have local state. Up until you said anything, I thought everything in kinto-admin *did* use Redux/Sagas. Now I'm not sure -- I guess anything with a `setState()` call is skirting it. I feel like hewing closer to Redux/Sagas is probably better than hewing further away from it, but I guess Redux/Saga probably also imposes a certain amount of boilerplate and that is unpleasant? I guess I feel like maybe avoiding Redux/Sagas in some cases is OK, but probably talking to servers should be done using Redux/Sagas.

leplatrem:
> . it seems that most of the state is handled in the component when it is local (like the auth methods in the AuthForm I guess?).

Yes, if I remember well what @n1k0 told me, we use the local state only for minor things like toggles, list expansion etc. 

> In a project using redux/sagas for everything I would have done it, in this specific case, I'm not sure... 

I would say that kinto-admin uses sagas for everything. If you spotted a place that makes this statement partly incorrect, then please open an issue we can track it as tech-debt :)

> Redux/Saga probably also imposes a certain amount of boilerplate and that is unpleasant?

Oh yeah indeed. It also requires to «name» many things like actions, global state attributes etc. But in the end, if it's consistent with the rest, it make sense

>  talking to servers should be done using Redux/Sagas

Yes, agree. And also for important side effects like calling `setClient()`

magopian:
I moved the code to a saga, which will be easier to test (I'll push a commit soon).
The current implementation still has issues (need to remove all reference to `authMethods` in the `AuthForm` local state for example, and from the `SettingsState`, and fix some undefined issues, I love JS).

magopian:
Ok, I think this is now ready to be merged. There's one caveat left: whenever you make a change in the server input, it'll try to fetch the server info.

This is not optimal, but mitigated by the fact that the kinto-http.js library won't do a fetch unless the url "looks good" (eg: you have the url schema, the version etc...). Also, if you just select the server from the history, this won't be an issue.

We might want to fix that though in the future, and at the same time fix the fact that whenever there's a change event, the cursor is placed at the end of the input (if you try to add a few letters in the middle of the url, you'll have a hard time because after each keypress the cursor is moved to the end of the string).

@leplatrem @glasserc what do you think, is that good enough to be merged now?

leplatrem:
Quick reply regarding the onchange event, you can just debounce the event, and fetch the server info only when no character was typed for like 500ms for example

src/components/AuthForm.js
```diff
@@ -6,6 +6,9 @@ import React, { PureComponent } from "react";
 import BaseForm from "./BaseForm";
 import { omit } from "../utils";
 
+const defaultKintoServer = "https://kinto.dev.mozaws.net/v1/";
+const defaultAuth = "anonymous";
```
leplatrem: I think this constant could be named «anonymous auth» explicitly

src/components/AuthForm.js
```diff
@@ -363,18 +366,49 @@ export default class AuthForm extends PureComponent<
...
+    const { session: { serverInfo: { capabilities } } } = this.props;
+    // Check which of our known auth implementations are supported by the server.
+    const knownAuthMethods = [
+      "basicauth",
+      "account",
+      "fxa",
+      "ldap",
+      "portier",
+      "anonymous",
+    ];
```
leplatrem: nitpick: could be const on top of source file

src/components/AuthForm.js
```diff
@@ -363,18 +366,49 @@ export default class AuthForm extends PureComponent<
...
+    let supportedAuthMethods = [defaultAuth];
+    knownAuthMethods.forEach(authMethod => {
+      if (authMethod in capabilities) {
+        supportedAuthMethods.push(authMethod);
+      }
+    });
```
leplatrem: nitpick: `knownAuthMethods.filter(a => a in capabilities)` ?

src/components/AuthForm.js
```diff
@@ -363,18 +366,49 @@ export default class AuthForm extends PureComponent<
...
+  getSupportedAuthMethods = () => {
+    const { session: { serverInfo: { capabilities } } } = this.props;
+    // Check which of our known auth implementations are supported by the server.
+    const knownAuthMethods = [
+      "basicauth",
+      "account",
+      "fxa",
+      "ldap",
+      "portier",
+      "anonymous",
```
leplatrem: Note: There is not `anonymous` capability (can be removed from list IMO)

src/components/AuthForm.js
```diff
@@ -6,6 +6,9 @@ import React, { PureComponent } from "react";
 import BaseForm from "./BaseForm";
 import { omit } from "../utils";
 
+const defaultKintoServer = "https://kinto.dev.mozaws.net/v1/";
```
leplatrem: nitpick: uppercase variables names?

src/components/AuthForm.js
```diff
@@ -363,18 +366,49 @@ export default class AuthForm extends PureComponent<
...
+    return supportedAuthMethods;
+  };
+
   onChange = ({ formData }: { formData: Object }) => {
-    const { authType } = formData;
+    const { authType, server } = formData;
+    if (this.state.formData.server !== server) {
+      // Server changed, request its capabilities to check what auth methods it
+      // supports.
+      const anonymousAuth = { authType: "anonymous", server: server };
```
leplatrem: could use const anonymous type maybe

src/sagas/session.js
```diff
@@ -15,25 +15,46 @@ import { saveSession, clearSession } from "../store/localStore";
...
     const serverInfo = yield call([client, client.fetchServerInfo]);
+    // Notify they're received
+    yield put(actions.serverInfoSuccess(serverInfo));
+
+    yield put(notificationActions.clearNotifications({ force: true }));
+  } catch (error) {
+    // Reset the server info that we might have added previously to the state.
+    yield put(actions.serverInfoSuccess(DEFAULT_SERVERINFO));
+    yield put(
+      notificationActions.notifyError("Couldn't retrieve server info", error)
```
leplatrem: `Could not reach server` maybe?

test/sagas/session_test.js
```diff
@@ -37,21 +39,77 @@ describe("session sagas", () => {
...
+        expect(getServerInfo.next().value).eql(
+          put(notificationsActions.clearNotifications({ force: true }))
+        );
+      });
     });
 
-    it("should configure the client", () => {
-      expect(getClient().remote).eql(authData.server);
+    describe("Failure", () => {
+      it("should reset the server info", () => {
```
leplatrem: just out of curiosity, why should we reset server info?

magopian: If you enter a bogus server in the input box, you don't want to still have the capabilities from the previous "working" server. Does that answer your question? Did I understand your question correctly?

src/components/AuthForm.js
magopian: What do you mean?

src/components/AuthForm.js
magopian: And then not forget to add the "anonymous" method which isn't in the capabilities as you said. Looks like a good idea, thanks ;)

test/sagas/session_test.js
leplatrem: yep! got it, wasn't obvious from the reviewer standpoint :)

magopian: Good point, I'll add a comment in the code ;) Thanks for bringing it up!

src/components/AuthForm.js
leplatrem: Something like `const ANONYMOUS_AUTH = "anonymous"` ... but that's kind of redundant with the other comments regarding this anonymous string

magopian: Ok, I added a ANONYMOUS_AUTH_DATA utility.

magopian:
@leplatrem I believe I've addressed your comments, what do you think?

PR #485, opened 2018-03-20
Debounce the AuthForm.onChange event

### LINKED ISSUE #484: Debounce the `AuthForm.onChange` event
Following #482 make sure that the `onChange` doesn't trigger a `getServerInfo` too often.

magopian:
Fixes #484

### DIFF
--- a/src/components/AuthForm.js
+++ b/src/components/AuthForm.js
@@ -4,7 +4,7 @@ import type { SessionState, SettingsState } from "../types";
 import React, { PureComponent } from "react";
 
 import BaseForm from "./BaseForm";
-import { omit } from "../utils";
+import { debounce, omit } from "../utils";
 
 const DEFAULT_KINTO_SERVER = "https://kinto.dev.mozaws.net/v1/";
 const ANONYMOUS_AUTH = "anonymous";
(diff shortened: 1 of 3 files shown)


src/components/AuthForm.js
```diff
@@ -14,16 +14,26 @@ const anonymousAuthData = server => ({
 });
 const KNOWN_AUTH_METHODS = ["basicauth", "account", "fxa", "ldap", "portier"];
 
+function debounce(fn, delay) {
```
leplatrem: It could go to `../utils`  IMO

src/components/AuthForm.js
```diff
@@ -14,16 +14,26 @@ const anonymousAuthData = server => ({
...
+  var timer = null;
+  return function() {
+    var context = this,
+      args = arguments;
+    clearTimeout(timer);
+    timer = setTimeout(() => fn.apply(context, args), delay);
+  };
+}
```
leplatrem: **nitpick**:
I have the impression that you could remove the context closure with a short array function no? (since this is maintained)
```js
function debounce(fn, delay) {
  var timer = null;
  return (...args) => {
    clearTimeout(timer);
    timer = setTimeout(() => fn.apply(this, args), delay);
  };
}
```

magopian: I believe this would work, indeed, I'll give it a try.

You mean
```javascript
timer = setTimeout(() => fn(args), delay);
```

Right?

magopian:
Thanks for the review @leplatrem I believe this is now better, what do you think?

PR #493, opened 2018-03-26
Reset the server capabilities straight away on server change

### LINKED ISSUE #487: auth methods from server capabilities aren't reset straight away when changing servers
When changing servers on the auth form, the list of supported auth methods isn't reset before the new capabilities are received, see https://github.com/Kinto/kinto-admin/blob/cf6c7b2656f3e7eca34d23e4e15ffcd01c9ce9e7/src/sagas/session.js#L33

Instead, as soon as we select a new server, the list of supported auth methods should be reset to listing only the "anonymous" method.

magopian:
Fixes #487


leplatrem:
Does it also reset errors in the UI? (like unreachable server etc.)

magopian:
It didn't, but it does now, thanks for pointing that out @leplatrem!

Please let me know if you're happy with the code now and I'll merge ;)

PR #496, opened 2018-03-27
Do not try to fetch server if not a proper URL

leplatrem:
This little test prevents trying to fetch malformed server URLs, which can happen when typing/editing the field value

@magopian if you agree with the idea, I can add tests etc


magopian:
I could be wrong, but kinto.js already does that, at least partially? For example it doesn't send a request if the version is missing, of the schema...

leplatrem:
Not directly:

https://github.com/Kinto/kinto-http.js/blob/d568f5b89854345fa39065e1b348b17788273383/src/base.js#L51-L56

https://github.com/Kinto/kinto-http.js/blob/d568f5b89854345fa39065e1b348b17788273383/src/base.js#L109-L113

Plus, it would come up as an error, while the user might just not be finished to enter the value (eg. `http://serv`)

leplatrem:
> I'm not sure this is good as it is now, regarding UX: what if you type htps://foo.bar/v1?

Yeah you're right this is bad.

What if we just simply set the `pattern` attribute on the text field? The field is going to be shown as red until a valid URL is written, and then only the fetch will kick in. I think it could work and would much simpler than a new state attribute

leplatrem:
I found and fixed other problems too: like for example, clicking on the auth type took the debounce time (500ms) to show in the credentials form

leplatrem:
Hehehe, thank you :heart:

leplatrem:
My CSS change affects the whole UI, I wonder if this is something we want, so I'll reduce the scope of the CSS selector to the AuthForm and open an issue :) 

screenshot from 2018-03-28 09-51-07

PR #497, opened 2018-03-27
Fix #488: Prevent race conditions when changing server URL

### LINKED ISSUE #488: Race condition when getting server capabilities for the list of auth methods
It's possible to receive the server info from a laggy server that was requested previously, AFTER the current server has been requested and answered

The scenario is the following:
1/ enter the address for LAGGY_SERVER (or select it from the history dropdown)
2/ before LAGGY_SERVER answers to the request for capabilities, change the address to FASTER_SERVER
3/ the capabilities from FAST_SERVER are received, and the list of supported auth methods is displayed
4/ LAGGY_SERVER answers with its list of supported auth methods, which will override the list of supported auth methods for the current server displayed (FAST_SERVER)

Not sure what's the best way to go about that, maybe tag a `getServerInfo` request with the server, so when we receive an answer, we make sure it's for the currently selected one?

leplatrem:
Fix #488

My proposal: use the global `client` to make sure that the response of fetchServerInfo concerns the latest server change

I tested it locally and it works nicely. Using `https://localhost:8888/v1` takes like 20 secondes to timeout so it's a perfect test case :)

@magopian if you're happy with the approach, then I can polish with tests etc.


src/sagas/session.js
```diff
@@ -24,16 +24,30 @@ export function* getServerInfo(
...
+  setupClient(auth);
+  // Fetch server information
+  const client = getClient();
   try {
-    setupClient(auth);
-    // Fetch server information
-    const client = getClient();
     const serverInfo = yield call([client, client.fetchServerInfo]);
+
+    // Check that the client was not changed in the mean time.
```
magopian: Maybe add something a bit more explicative? Something like:
```
// Check that the client was not changed in the mean time. This could happen if a request to
// another server was sent after the current one, but took less time to answer.
// In such a case, this means this saga/server request should be discarded in favor of the other one
// which was sent later.
```

PR #505, opened 2018-03-29
Fixup #496: restore #493 fix

magopian:
Fixup: #496 reverted most of what was fixed in #493.

This code is currently "broken": selecting a server from the history dropdown
shouldn't be debounced, but should instead reset the server capabilities (and
notifications) straight away, and send the request for the capabilities.

### DIFF
--- a/src/components/AuthForm.js
+++ b/src/components/AuthForm.js
@@ -40,6 +40,7 @@ class ServerHistory extends PureComponent<
       event.preventDefault();
       this.props.onChange(server);
       this.setState({ menuOpened: false });
+      this.debouncedFetchServerInfo(server);
     };
   };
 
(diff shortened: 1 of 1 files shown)


src/components/AuthForm.js
```diff
@@ -40,6 +40,7 @@ class ServerHistory extends PureComponent<
       event.preventDefault();
       this.props.onChange(server);
       this.setState({ menuOpened: false });
+      this.debouncedFetchServerInfo(server);
```
magopian: This should really be `this.fetchServerInfo(server)` because we don't want to debounce (and wait 500ms) when selecting a server from the history dropdown. However, I couldn't understand why, when doing that, the value of the input doesn't change.

It seems that by using `fetchServerInfo` without debouncing has the consequence of calling `AuthForm.onChange` with the current server (and not the new one). This is a total mystery for me, and might be a bug in the react-json-schema-form library (or maybe not).

Please @glasserc @leplatrem if you have any idea what's causing that, let me know.

leplatrem: I think I ran into something similar when working on #496 

I'll look into it :)

magopian: Thanks a lot, and if you find out what the issue is, please let me know how you found about it! I tried many different ways to investigate, nothing would make sense...

glasserc: That sounds a little bit like https://github.com/mozilla-services/react-jsonschema-form/issues/446...

leplatrem:
From the user perspective, the behaviour of this branch is good. 
Plus, debouncing the select change event is not completely irrelevant. If the user scrolls in the combbox, you don't want to fire hundreds of requests...

So r+ :)

PR #509, opened 2018-03-30
Fix #460: Add OpenID connect support

### LINKED ISSUE #460: Add OpenID connect support
* Read identity providers from OpenID capability (root URL)
* Fetch `${idp}/.well-known/openid-configuration`
* Start redirection to do login «dance»
* Read userinfo from ID token (display name top right menu)
* Put access token in session storage
* Use access token in `Authorization: Bearer {}` requests headers

Depends on https://github.com/Kinto/kinto/pull/1425#discussion_r165250026

magopian:
Fixes #460

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,6 @@ The `KintoAdmin` component accepts a `settings` prop, where you can define the f
 
 - `maxPerPage`: The max number of results per page in lists (default: `200`).
 - `singleServer`: The server URL to be used (default: `null`). This removes the ability to connect to multiple servers.
-- `authMethods`: The list of supported authenticated methods (default: `["basicauth", "ldap", "fxa", "portier", "openid"]`). The first is the default one.
 - `sidebarMaxListedCollections`: The maximum number of bucket collections entries to list in the sidebar.
 
 Example:
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ The `KintoAdmin` component accepts a `settings` prop, where you can define the f
 
 - `maxPerPage`: The max number of results per page in lists (default: `200`).
 - `singleServer`: The server URL to be used (default: `null`). This removes the ability to connect to multiple servers.
-- `authMethods`: The list of supported authenticated methods (default: `["basicauth", "ldap", "fxa"]`). The first is the default one.
+- `authMethods`: The list of supported authenticated methods (default: `["basicauth", "ldap", "fxa", "portier", "openid"]`). The first is the default one.
 - `sidebarMaxListedCollections`: The maximum number of bucket collections entries to list in the sidebar.
 
 Example:
--- a/src/actions/session.js
+++ b/src/actions/session.js
@@ -162,13 +162,11 @@ export function navigateToOpenID(
   const { origin, pathname } = document.location;
   const { server } = authFormData;
   const strippedServer = server.replace(/\/$/, "");
-  const { auth_path } = provider;
-  const strippedAuthPath = auth_path.replace(/^\//, "");
+  const { auth_path: authPath } = provider;
+  const strippedAuthPath = authPath.replace(/^\//, "");
   const payload = btoa(JSON.stringify(authFormData));
-  const redirect = `${origin}${pathname}#/auth/${payload}/`;
-  document.location.href = `${strippedServer}/${strippedAuthPath}?callback=${encodeURIComponent(
-    redirect
-  )}&scope=openid email`;
+  const redirect = encodeURIComponent(`${origin}${pathname}#/auth/${payload}/`);
+  document.location.href = `${strippedServer}/${strippedAuthPath}?callback=${redirect}&scope=openid email`;
   return { type: null };
 }
 
--- a/src/actions/session.js
+++ b/src/actions/session.js
@@ -155,14 +155,20 @@ function postToPortier(server: string, redirect: string): NavigationResult {
   }
 }
 
-function navigateToOpenID(
-  server: string,
-  provider: string,
-  redirect: string
+export function navigateToOpenID(
+  authFormData: Object,
+  provider: Object
 ): NavigationResult {
-  document.location.href = `${server}openid/${provider}/login?redirect=${encodeURIComponent(
+  const { origin, pathname } = document.location;
+  const { server } = authFormData;
+  const strippedServer = server.replace(/\/$/, "");
+  const { auth_path } = provider;
+  const strippedAuthPath = auth_path.replace(/^\//, "");
+  const payload = btoa(JSON.stringify(authFormData));
+  const redirect = `${origin}${pathname}#/auth/${payload}/`;
+  document.location.href = `${strippedServer}/${strippedAuthPath}?callback=${encodeURIComponent(
     redirect
-  )}`;
+  )}&scope=openid email`;
   return { type: null };
 }
 
@@ -180,9 +186,6 @@ export function navigateToExternalAuth(authFormData: Object): NavigationResult {
       return navigateToFxA(server, redirect);
     } else if (authType === "portier") {
       return postToPortier(server, redirect);
-    } else if (authType === "openid") {
-      const { provider } = authFormData;
-      return navigateToOpenID(server, provider, redirect);
     } else {
       return notifyError(`Unsupported auth navigation type "${authType}".`);
     }
--- a/src/actions/session.js
+++ b/src/actions/session.js
@@ -155,6 +155,17 @@ function postToPortier(server: string, redirect: string): NavigationResult {
   }
 }
 
+function navigateToOpenID(
+  server: string,
+  provider: string,
+  redirect: string
+): NavigationResult {
+  document.location.href = `${server}openid/${provider}/login?redirect=${encodeURIComponent(
+    redirect
+  )}`;
+  return { type: null };
+}
+
 /**
  * Massive side effect: this will navigate away from the current page to perform
  * authentication to a third-party service, like FxA.
@@ -169,6 +180,10 @@ export function navigateToExternalAuth(authFormData: Object): NavigationResult {
       return navigateToFxA(server, redirect);
     } else if (authType === "portier") {
       return postToPortier(server, redirect);
+    } else if (authType === "openid") {
+      // TODO: const { provider } = authFormData;
+      const provider = "auth0";
+      return navigateToOpenID(server, provider, redirect);
     } else {
       return notifyError(`Unsupported auth navigation type "${authType}".`);
     }
--- a/src/actions/session.js
+++ b/src/actions/session.js
@@ -181,8 +181,7 @@ export function navigateToExternalAuth(authFormData: Object): NavigationResult {
     } else if (authType === "portier") {
       return postToPortier(server, redirect);
     } else if (authType === "openid") {
-      // TODO: const { provider } = authFormData;
-      const provider = "auth0";
+      const { provider } = authFormData;
       return navigateToOpenID(server, provider, redirect);
     } else {
       return notifyError(`Unsupported auth navigation type "${authType}".`);
--- a/src/client.js
+++ b/src/client.js
@@ -8,7 +8,8 @@ let client: ?KintoClient;
 
 function getAuthHeader(auth: AuthData): ?string {
   switch (auth.authType) {
-    case "fxa": {
+    case "fxa":
+    case "openid": {
       const { token }: { token: string } = auth.credentials;
       return "Bearer " + token;
     }
--- a/src/components/AuthForm.js
+++ b/src/components/AuthForm.js
@@ -310,7 +310,7 @@ const authSchemas = authType => {
   };
 };
 
-const authLabels = authType => {
+const getAuthLabel = authType => {
   const labels = {
     anonymous: "Anonymous",
     basicauth: "Basic Auth",
@@ -320,10 +320,11 @@ const authLabels = authType => {
     portier: "Portier",
   };
   if (authType.startsWith("openid-")) {
+    // The authType openid-<provider> is constructed in getSupportedAuthMethods.
     const provider = authType.replace("openid-", "");
     const prettyProvider =
       labels[provider] || provider.charAt(0).toUpperCase() + provider.slice(1);
-    return `${prettyProvider} (OpenID)`;
+    return `OpenID Connect (${prettyProvider})`;
   }
   return labels[authType];
 };
@@ -340,7 +341,7 @@ function extendSchemaWithHistory(schema, history, authMethods, singleServer) {
       authType: {
         ...schema.properties.authType,
         enum: authMethods,
-        enumNames: authMethods.map(authLabels),
+        enumNames: authMethods.map(getAuthLabel),
       },
       server: {
         ...schema.properties.server,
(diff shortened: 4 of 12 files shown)


magopian:
The looks will have to be improved: I wanted to have multiple submit buttons (one for each openID provider), but that's not possible with react-json-schema-form. Well, you can have multiple submit buttons, but you won't know which one was pressed (the `value` isn't passed down after the validation it seems).

For now, this is how it looks:

screen shot 2018-03-30 at 16 40 43

No "auth0 dance" is implemented yet.

leplatrem:
Cool, that's a start!

In our use case we'll have only one provider, so for that particular case we case even hide the radio

src/client.js
```diff
@@ -8,7 +8,8 @@ let client: ?KintoClient;
 
 function getAuthHeader(auth: AuthData): ?string {
   switch (auth.authType) {
-    case "fxa": {
+    case "fxa":
+    case "openid": {
```
leplatrem: Ideally the `Bearer` string should be taken from the capability

magopian:
This currently doesn't work, check the TODOs in the code.
Multiple issues:
1/ we overwrite the "server info" from the authentified request we do when being redirected after an auth0 login dance, with the "server info" from the default server that is displayed when we mount the AuthForm component
2/ when we get the user id back from the server, it looks like `<provider>:<token>`. But what we're expecting is `<authType>:<token>`. So we receive eg `auth0:token` instead of `openid:token`, in `setupSession`

leplatrem:
I was thinking of something: instead of having two list of bullets points, we could merge the OpenID method and the list of provider into a single list, like this:
- Anonymous
- Basic Auth
- Google (OpenID)
- Auth0 (OpenID)
- FxA (OpenID)

magopian:
So, the current state of this PR is: it should be working, from what I've tested locally with a local kinto server using the configuration from https://github.com/leplatrem/kinto-oidc-demo#configure--run-kinto, and the google openid provider.

There's a known caveat though: when logging out, the server capabilities aren't fetched again (and are removed from the session), so the supported auth methods aren't displayed until the server is changed or the page is refreshed. This is because it's a nightmare of race conditions and I can't find where and when to get the server information in order to cover all the edge cases.

There's still a bunch of debug messages (`console.log`) littered in the code, and probably some missing tests, but it's in a state where it should be testable. I also haven't addressed your comment https://github.com/Kinto/kinto-admin/pull/509#discussion_r178312847 yet.

What do you think @leplatrem ?

magopian:
I added a test, and also implemented your comment about having one auth method per provider "inlined":

screen shot 2018-05-18 at 10 25 12

leplatrem:
Nice!

I will look at the code in details next week.

@peterbe could you try it with your Kinto instance configured with Auth0?

Regarding the logout bug, it's not a critical failure. We could ship the feature and mark it as *experimental* while we work on it. It would allow us to slowly go forward with the OpenID deployment for remote settings, and more importantly, take over what you started to give you some relief :)

peterbe:
Who the heck logs out anyway? The only time I ever log out is my online bank and when I realize the mistake of logging in with peterbe@m.c when I should have logged in with pbengtsson@m-c. :)

peterbe:
screen shot 2018-05-18 at 10 27 44 am

Right after the redirect from Auth0.com's log in screen, there was a page with lots of ugly debugging info near the top of the page. Then it redirected to what it looks like in this screenshot.

leplatrem:
> Then it redirected to what it looks like in this screenshot.

:tada:

magopian:
I'm not sure I understand your comment @peterbe? Is there something wrong in the screenshot you see?
From what I see, everything seems to have worked correctly. The debug output is probably linked to my comment in https://github.com/Kinto/kinto-admin/pull/509#issuecomment-389901873 : there's still a bunch of `console.log` that's left in the code while I make sure this feature is working as expected.

Once you folks are happy with what's here, I'll clean it up ;)

peterbe:
> Is there something wrong in the screenshot you see?

Half a second before this screenshot, right after being redirected back from Auth0 the screen looked very different. It flashed by so quickly I couldn't really see what it said. It looked like a piece of JSON being displayed in the blue flash message near the top. I'll try to make it happen again. 
I'm not referring to `console.log` debug messages. It was in the DOM.

peterbe:
@magopian Actually, instead of me trying to reproduce it, and probably failing to make a screenshot, how about I send you my kinto.ini (with my personal Auth0 credentials) to you and you can try it yourself locally?

magopian:
I think I can now reproduce this issue: when we get back from the auth0 dance, we land on an url like

http://localhost:3000/#/auth/<token payload>

And the payload is displayed in the breadcrumbs for some reason, and appears like that:

screen shot 2018-05-22 at 15 06 38


Is that what you saw @peterbe ?

magopian:
@peterbe ok I believe I've fixed the issue: I've excluded the `auth` crumb in the breadcrumbs.

peterbe:
Yes! That's what my poor eyes had to suffer through.

magopian:
ok great, so I'll just clean the PR, see if I can add a couple more tests maybe, and ask you if you're happy with me merging it 👍

README.md
```diff
@@ -81,7 +81,7 @@ The `KintoAdmin` component accepts a `settings` prop, where you can define the f
 
 - `maxPerPage`: The max number of results per page in lists (default: `200`).
 - `singleServer`: The server URL to be used (default: `null`). This removes the ability to connect to multiple servers.
-- `authMethods`: The list of supported authenticated methods (default: `["basicauth", "ldap", "fxa"]`). The first is the default one.
+- `authMethods`: The list of supported authenticated methods (default: `["basicauth", "ldap", "fxa", "portier", "openid"]`). The first is the default one.
```
leplatrem: Is `authMethods` still a thing?

src/actions/session.js
```diff
@@ -155,6 +155,23 @@ function postToPortier(server: string, redirect: string): NavigationResult {
...
+  const { origin, pathname } = document.location;
+  const { server } = authFormData;
+  const strippedServer = server.replace(/\/$/, "");
+  const { auth_path } = provider;
+  const strippedAuthPath = auth_path.replace(/^\//, "");
+  const payload = btoa(JSON.stringify(authFormData));
+  const redirect = `${origin}${pathname}#/auth/${payload}/`;
+  document.location.href = `${strippedServer}/${strippedAuthPath}?callback=${encodeURIComponent(
+    redirect
+  )}&scope=openid email`;
```
leplatrem: nit: maybe use intermediary variables for callback and scope to avoid wrapping?

src/components/AuthForm.js
```diff
@@ -178,133 +185,147 @@ const loginPasswordUiSchema = {
...
+  const labels = {
+    anonymous: "Anonymous",
+    basicauth: "Basic Auth",
+    account: "Kinto Account Auth",
+    fxa: "Firefox Account",
+    ldap: "LDAP",
+    portier: "Portier",
+  };
+  if (authType.startsWith("openid-")) {
+    const provider = authType.replace("openid-", "");
```
leplatrem: Maybe add a comment here that we construct this `openid-*` string in `getSupportedAuthMethods()`

src/components/AuthForm.js
```diff
@@ -442,15 +468,40 @@ export default class AuthForm extends PureComponent<
...
+        const {
+          serverInfo: {
+            capabilities: { openid: { providers } = { providers: [] } },
+          },
+        } = session;
+        const providerData = providers.find(p => p.name === openidProvider);
+        if (!providerData) {
+          throw "Couldn't find provider data in the state. Bad.";
+        }
+        console.log("providerData", providerData);
```
leplatrem: leftover or intentional info?

src/components/AuthForm.js
```diff
@@ -442,15 +468,40 @@ export default class AuthForm extends PureComponent<
...
+      case "openid": {
+        const { session } = this.props;
+        const {
+          serverInfo: {
+            capabilities: { openid: { providers } = { providers: [] } },
+          },
+        } = session;
+        const providerData = providers.find(p => p.name === openidProvider);
+        if (!providerData) {
+          throw "Couldn't find provider data in the state. Bad.";
```
leplatrem: use `throw new Error(str)`

src/sagas/session.js
```diff
@@ -84,9 +92,16 @@ export function* setupSession(
     const { authType } = auth;
     if (
       authType != "anonymous" &&
-      (!userId || !userId.startsWith(authType + ":"))
+      (!userId ||
+        (!userId.startsWith(authType + ":") &&
+          // If the authType is openid, the userId doesn't start with "openid:" but "auth0:".
+          (authType.startsWith("openid-") && !userId.startsWith("auth0:"))))
```
leplatrem: I don't understand this test.

The user id starts with the provider name (the right part of authType if I understand correctly)

test/sagas/session_test.js
```diff
@@ -207,6 +233,31 @@ describe("session sagas", () => {
...
+      it("should correctly authenticate the user when using openID", () => {
+        const authData = {
+          server: "http://server.test/v1",
+          authType: "openid-google",
+          credentials: { token: "the token" },
+        };
+        const serverInfo = {
+          ...serverInfo,
+          user: { id: "auth0:token" },
+        };
```
leplatrem: I would say this situation is impossible to obtain with a real server.

If you authenticate with a provider name `google`, then your userid will be `google:tartempion`

magopian: are you sure about that? In my tests, i got `auth0:...` back, which was a surprise to me!

src/sagas/session.js
magopian: are you sure about that? In my tests, i got `auth0:...` back, which was a surprise to me!

src/actions/session.js
peterbe: I think it's Pretty :)

src/actions/session.js
```diff
@@ -155,6 +155,23 @@ function postToPortier(server: string, redirect: string): NavigationResult {
...
 }
 
+export function navigateToOpenID(
+  authFormData: Object,
+  provider: Object
+): NavigationResult {
+  const { origin, pathname } = document.location;
+  const { server } = authFormData;
+  const strippedServer = server.replace(/\/$/, "");
+  const { auth_path } = provider;
```
peterbe: nit: This one's under_score but the next one is camelCase. I.e. `auth_path` and `[stripped]AuthPath`.

src/components/AuthForm.js
```diff
@@ -178,133 +185,147 @@ const loginPasswordUiSchema = {
...
 };
 
-const authLabels = {
-  anonymous: "Anonymous",
-  basicauth: "Basic Auth",
-  account: "Kinto Account Auth",
-  fxa: "Firefox Account",
-  ldap: "LDAP",
-  portier: "Portier",
+const authLabels = authType => {
```
peterbe: Shouldn't this be called `authLabel` (or `getAuthLabel`)?

src/components/AuthForm.js
```diff
@@ -178,133 +185,147 @@ const loginPasswordUiSchema = {
...
+    account: "Kinto Account Auth",
+    fxa: "Firefox Account",
+    ldap: "LDAP",
+    portier: "Portier",
+  };
+  if (authType.startsWith("openid-")) {
+    const provider = authType.replace("openid-", "");
+    const prettyProvider =
+      labels[provider] || provider.charAt(0).toUpperCase() + provider.slice(1);
+    return `${prettyProvider} (OpenID)`;
```
peterbe: If `authType` is something like `openid-Auth0.com` (aka `openid-${provider.name}`) then the name *is* `authType.replace("openid-", "");`

I think it should be the other way around. I.e.
```javascript
const providerName = authType.replace("openid-", "");
return `OpenID Connect (${provider})`
```

*Note!* It's called "OpenID Connect", not "OpenID". "OpenID" was some other junk we tried and failed 10-15 years ago. "OpenID Connect" is a OAuth2 sub-standard.

src/components/AuthForm.js
```diff
@@ -414,24 +436,28 @@ export default class AuthForm extends PureComponent<
...
   };
 
   onChange = ({ formData }: { formData: Object }) => {
     const { authType } = formData;
-    const { schema, uiSchema } = authSchemas[authType];
-    const specificFormData = [ANONYMOUS_AUTH, "fxa", "portier"].includes(
-      authType
-    )
+    const { uiSchema } = authSchemas(authType);
+    let { schema } = authSchemas(authType);
```
peterbe: Why `let` and not `const`? I don't see it changing.

peterbe:
screen shot 2018-05-23 at 8 06 24 am

Nice that it caught my typo and handled the Network error gracefully.

peterbe:
By the way, I re-tested actually logging in using my local server and it still works. 
I must admit that I haven't really the discussion about the prefix on the email. Like, why am I logged in as `auth0:peterbe@mozilla.com` and not `peterbe@mozilla.com`?

peterbe:
When I click "Logout", I get logged out but now the known providers are lost:

screen shot 2018-05-23 at 8 12 24 am

I have to put my cursor into the Server field, make a change (e.g. editing it to something different and editing back to what it should be) and then the two options appear again.

peterbe:
> If that something you'd like to change or optionally disable, we should discuss it there :)

That got more complicated than I had hoped. I guess if kinto-admin is for techy super users it's OK to "butcher" my email address. All I know is that my email is `peterbe@mozilla.com` not `auth0:peterbe@mozilla.com`. But I can see the value in including info about how I logged in. Especially since it plants in my head that the next time I log in, I can remember that the last time I used OpenID Connect and not FxA.

leplatrem:
> I guess if kinto-admin is for techy super users it's OK to "butcher" my email address. All I know is that my email is peterbe@mozilla.com not auth0:peterbe@mozilla.com.

`auth0:peterbe@mozilla.com` is not your email address. It's your user id.

And BTW, your user id looks like an email address because you configured your server [in some specific way](https://kinto.readthedocs.io/en/stable/configuration/settings.html#openid-connect).

But let's discuss that elsewhere :)

peterbe:
> auth0:peterbe@mozilla.com is not your email address. It's your user id.

I know I know :) I was just playing devil's advocate. It just looks a bit odd when so many other (consumer) websites display my email address in a regular way. Granted, so many other (consumer) websites display it as "Peter Bengtsson" instead :)

src/sagas/session.js
magopian: seems my issue with the `auth0:` prefix was due to some configuration problem for the kinto running locally... I've fixed that (and the test)

magopian:
> When I click "Logout", I get logged out but now the known providers are lost:
> 
> ...
> 
> I have to put my cursor into the Server field, make a change (e.g. editing it to something different and editing back to what it should be) and then the two options appear again.

@peterbe this is what I was referring as the caveat in https://github.com/Kinto/kinto-admin/pull/509#issuecomment-389901873

I believe I've fixed the nits and renamed `Google (openID)` to `OpenID Connect (google)`.

src/client.js
leplatrem: @magopian can you open a follow-up issue for this please?

magopian: opened #543

magopian:
Should I merge this? Is this happening?

PR #510, opened 2018-04-06
Show details in popup in HistoryTable component

### LINKED ISSUE #492: History table is not displayed correctly
screenshot from 2018-03-24 13-56-11

stloma:
Fixes #492 

An element in two nested `<thead>` tags was being returned. The intention was to have a table row hide/show under the current row when details were exposed. As far as I know, with the way it's written, it is hard to do pre-fragments (React versions < 16) without some in depth rewrites.

My idea was just to take the element out of the document flow.

screenshot-2018-4-6 kinto web administration dev

### DIFF
--- a/css/styles.css
+++ b/css/styles.css
@@ -197,6 +197,10 @@ table.record-list td.load-more {
   line-height: 2.5em;
 }
 
+table.record-list > tbody:nth-child(odd) tr td {
+  background-color: #fff;
+}
+
 .load-more > a:hover, .load-more > a:active {
   background: #d9edf7;
   text-decoration: none;
--- a/src/components/HistoryTable.js
+++ b/src/components/HistoryTable.js
@@ -152,47 +152,53 @@ class HistoryRow extends PureComponent<HistoryRowProps, HistoryRowState> {
     } = target;
 
     return (
-      <tr>
-        <td>
-          <span title={humanDate(last_modified)}>{timeago(last_modified)}</span>
-        </td>
-        <td>{action}</td>
-        <td>{resource_name}</td>
-        <td>
-          <AdminLink
-            name={`${resource_name}:attributes`}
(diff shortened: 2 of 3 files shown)


css/styles.css
```diff
@@ -207,7 +207,11 @@ table.record-list td.load-more {
 }
 
 .history-row-details {
-  display: table-row;
+  text-align: left;
+  position: absolute;
+  right: 5em;
+  z-index: 99;
+  width: 250px;
```
leplatrem: Something like `max-width: 80%` would avoid horizontal scrolling in most cases

leplatrem:
For the reference, here is what I obtain on `master` when applying stloma's fix with the browser devtools

screenshot-2018-4-9 kinto web administration dev

stloma:
Ah, yes, sorry about the mess, got ahead of myself :)

Should be taken care of.

stloma:
I've actually been working on this and think I have a solution.

According to the HTML5 spec, you're allowed to use multiple `<tbody>` elements inside a `<table>`. Using that technique I can implement multiple `<tbody>` elements with nested `<tr>`s to dropdown a details row under the current row within the React spec.

In any case, just wanted to menion I think I firgured out how to do it without a popup, I'll be pushing an example soon for discussion.

stloma:
I think this solution will work. The nesting looks like this, which is valid:

```
<table>
  <thead>
    <tr>
      <th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>
    </tr>
    <tr>
      Details
    </tr>
  </tbody>
<table>

```

Other than this, I'm having trouble coming up with a solution that works without the popup, but am definitely open to thoughts on this solution vs. popup vs some other solution.

glasserc:
I think so too. Closing for now. Sorry @stloma. Feel free to reopen if we missed something.

PR #516, opened 2018-04-16
Fix log out on refresh (fixes #499)

### LINKED ISSUE #499: Refresh pages logs out
Probably due to #331, refreshing the page looses the current session (ie. logs the user out)

leplatrem:
fixes #499 

- [x] Fix bug
- [ ] Add tests?
- [ ] ~~Refactor session save/load?~~

### DIFF
--- a/test/setup-jsdom.js
+++ b/test/setup-jsdom.js
@@ -6,6 +6,9 @@ const JSDOM = new jsdom.JSDOM("<!doctype html><html><body></body></html>");
 global.window = JSDOM.window;
 global.document = window.document;
 global.navigator = global.window.navigator;
+// Fake a window.location to be a real url, and not `about:blank` which breaks
+// react-router-redux's `syncHistoryWithStore`, called from the KintoAdmin
+// component constructor.
 JSDOM.reconfigure({ url: "http://server.test/" });
 
 // Setup dumb sessionStorage for tests
--- a/test/setup-jsdom.js
+++ b/test/setup-jsdom.js
@@ -6,6 +6,7 @@ const JSDOM = new jsdom.JSDOM("<!doctype html><html><body></body></html>");
 global.window = JSDOM.window;
 global.document = window.document;
 global.navigator = global.window.navigator;
+JSDOM.reconfigure({ url: "http://server.test/" });
 
 // Setup dumb sessionStorage for tests
 global.sessionStorage = (function() {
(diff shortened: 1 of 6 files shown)


glasserc:
This is just a drive-by review, I'm not an expert in the linked code, etc.

It looks to me like the bug is that the `getServerInfo` calls happen in the AuthForm, "after" the `setupSession` calls. Is that right?

I'm not a React expert, but it feels a bit odd to me that a "component" should be responsible for setting up the session. Isn't this something that should happen in a saga?

leplatrem:
> It looks to me like the bug is that the getServerInfo calls happen in the AuthForm, "after" the setupSession calls. Is that right?

Before this PR:
- A call to `sessionActions.setup()` is made using the info found in `window.sessionstorage`
- The AuthForm makes a call to `getServerInfo` anonymously when it mounts to obtain the authentication methods of the currently selected server
- The second anonymous call erases the previous one

After this PR:
- The AuthForm reads `window.sesssionstorage`
- If previous session found, execute same action as "submit" button
- If not, calls `getServerInfo` anonymously

> I'm not a React expert, but it feels a bit odd to me that a "component" should be responsible for setting up the session. Isn't this something that should happen in a saga?

Me neither. And I also think we can do better :) Like let the auth form wait for the response of the anonymous `getServerInfo` if no session is found

leplatrem:
> Do you want to add a test to make sure it fixes the initial issue?

hmm, there was no test for the part I touched, I don't really know where to start... Ideally the changes to some existing tests would have been:
- the `HomePage` does not fetch server info on mount anymore
- the `KintoAdmin`  component now fetches infos from the default server anonymously if no previous session is found

magopian:
Ok @leplatrem I added a couple of basic tests, I'm not sure if that's what you had in mind? Do you want me to work on those some more, or do you want to take it from here?

test/setup-jsdom.js
```diff
@@ -6,6 +6,7 @@ const JSDOM = new jsdom.JSDOM("<!doctype html><html><body></body></html>");
 global.window = JSDOM.window;
 global.document = window.document;
 global.navigator = global.window.navigator;
+JSDOM.reconfigure({ url: "http://server.test/" });
```
leplatrem: Could you add a comment why you had to add this?

magopian: sure ;) It's to do with the `KintoAdmin` component calling `syncHistoryWithStore` which needs a real url (and not the base `about:blank` from jsdom)

magopian:
Should we merge this now?

glasserc:
I'm even less qualified to talk about it than you two but I would say yes.

leplatrem:
Thanks!

PR #523, opened 2018-04-20
Fix signoff workflow info (fixes #500)

### LINKED ISSUE #500: [signoff] Incoherent requested_review timestamps
screenshot-2018-3-27 kinto web administration

leplatrem:
Fixes #500

Branch based on #516 to preserve mental health while testing locally

screenshot from 2018-04-20 16-51-18
screenshot from 2018-04-20 16-56-33

* [x] Revamp whole code
* [ ] Fix tests


src/plugins/signoff/sagas.js
```diff
@@ -47,46 +48,78 @@ export function* onCollectionRecordsRequest(getState, action) {
...
     last_reviewer_comment: lastReviewerComment,
+    last_signature_by: lastSignatureBy,
+    last_signature_date: nLastSignatureDate,
   } = sourceAttributes;
+  let lastEditDate = Date.parse(nLastEditDate);
+  let lastReviewRequestDate = Date.parse(nLastReviewRequestDate);
+  let lastReviewDate = Date.parse(nLastReviewDate);
+  let lastSignatureDate = Date.parse(nLastSignatureDate);
+
+  // Retro-compatibility for old server versions.
```
glasserc: µnit: "Backwards compatibility"

src/plugins/signoff/sagas.js
```diff
@@ -47,46 +48,78 @@ export function* onCollectionRecordsRequest(getState, action) {
...
+  // This code could be removed after a few releases have been deployed to prod.
+  // (ie. no server rollback)
+  const {
+    capabilities: {
+      signer: { version },
+    },
+  } = serverInfo;
+  if (version < "3.2.0") {
+    // Some of these values will be wrong and disturbing in case of
+    // signature refresh.
```
glasserc: I love this use of the word "disturbing".

magopian:
Small note about the failing test: you might need `sinon.useFakeTimers`?

leplatrem:
> Small note about the failing test: you might need sinon.useFakeTimers?

The failure are due to inconsistent returned values of `Date.parse()`

`1523811083971−1523803883971` == 2H

Weird, because here it says explicitly that UTC will be used with ISO format:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse

leplatrem:
I think we made a mistake to put ISO date in those fields. I think there's still time to change to timestamps.

leplatrem:
Thanks for the detailed explanations :heart: 

> But I think supporting _since=2018-04-01T00:00:00 and _since=2018-04-01T00:00:00Z would be a good move in any case.

Not really IMO. I still dream of a day when our timestamp will be opaque (generated by a postgresql sequence for example). 
Plus, since those dates are generated in python independently from the storage timestamps, there is no guarantee that we'll obtain what we expect using them in `?_since=`.
I read the code again, and the only place we need to fetch a list with since is here:

https://github.com/Kinto/kinto-admin/blob/ae4232f60029e94a75414deee0b519f4fed6ed17/src/plugins/signoff/sagas.js#L114-L121

And it would a lot more correct to use the records plural endpoint Etag than the collection metadata timestamp (which changes at each signature refresh) or the last approval date (which may slightly skew from storage collection timestamp)

glasserc:
I agree that I would prefer to keep Etags and last_modifieds opaque. If we do, then there's no way to convert from the metadata timestamps to a `_since=` query, no matter what the wire format is. :+1:

PR #532, opened 2018-05-04
Fetch capabilities from singleServer if set

### LINKED ISSUE #531: Capabilities fetched from other server in single server mode
screenshot from 2018-05-04 11-57-05

screenshot from 2018-05-04 11-58-12

> Note: This also happens when session/cookies are cleaned...

@magopian can you look at it please?

It blocks deployment of kinto-dist 7.0.1 in PROD (https://bugzilla.mozilla.org/show_bug.cgi?id=1457150)

magopian:
Fixes #531

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -35,6 +35,7 @@ export default class KintoAdmin extends Component<Props> {
     const { plugins, settings } = props;
     this.store = configureStore({ settings }, plugins);
     syncHistoryWithStore(hashHistory, this.store);
+    const { history } = this.store.getState();
 
     // Restore saved session, if any
     const session = loadSession();
@@ -44,7 +45,10 @@ export default class KintoAdmin extends Component<Props> {
       this.store.dispatch(
         sessionActions.getServerInfo({
           authType: "anonymous",
-          server: DEFAULT_KINTO_SERVER,
+          server:
+            settings.singleServer ||
+            (history && history[0]) ||
+            DEFAULT_KINTO_SERVER,
         })
       );
     }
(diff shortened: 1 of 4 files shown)


test/index_test.js
```diff
@@ -28,10 +28,10 @@ describe("KintoAdmin", () => {
       buckets: [{}],
       serverInfo: {},
     };
-    const createKintoAdmin = () =>
+    const createKintoAdmin = (settings = {}) =>
       createComponent(KintoAdmin, {
         plugins: [],
-        settings: {},
+        settings: settings,
```
leplatrem: nit: with ES6 `{settings}` is enough

src/index.js
```diff
@@ -44,7 +45,10 @@ export default class KintoAdmin extends Component<Props> {
       this.store.dispatch(
         sessionActions.getServerInfo({
           authType: "anonymous",
-          server: DEFAULT_KINTO_SERVER,
+          server:
+            settings.singleServer ||
+            (history && history[0]) ||
```
leplatrem: Should we have a test for that too?

leplatrem: I'm just a bit puzzle that it looks very similar to https://github.com/Kinto/kinto-admin/blob/master/src/components/AuthForm.js#L315
I know that we don't want to refactor it now, but can we at least add a comment to «connect» the two places for future contributions?

test/index_test.js
magopian: you mean just `settings` right? Like so:

```javascript
    const createKintoAdmin = (settings = {}) =>
      createComponent(KintoAdmin, {
        plugins: [],
        settings,
      });
```

leplatrem: yes

src/index.js
magopian: I've just refactored it in `utils/getServerByPriority`

magopian:
@leplatrem I believe I've addressed all your comments, please let me know if there's anything else you'd like changed!

PR #536, opened 2018-05-19
Don't request records as part of the permissions

### LINKED ISSUE #533: Handling lots of permissions
Hi all, I have a kinto instance with lots of records. Since the permissions endpoint was made mandatory, kinto-admin now takes 15+ minutes to load 10s of pages of permissions from the permissions endpoint before it can be used. Even then it seems to run out of local storage before it completes.

I have paginate_by set to 1000, and probably 20k or so of records in there, total.

Is it possible to avoid loading permissions for every record in the database?
I'm I not supposed to have permissions for every record in kinto? Did I misconfigure something?

Please let me know if I'm asking this in the right place.

glasserc:
It's normal for a user to have a ton of records. We only actually need
the buckets and collections here to build the sidebar, but build this
as a blacklist instead of a whitelist on the off-chance that someday
we add groups to the sidebar too.

Fixes #533.

### DIFF
--- a/src/sagas/session.js
+++ b/src/sagas/session.js
@@ -229,7 +229,7 @@ export function* listBuckets(
     if ("permissions_endpoint" in serverCapabilities) {
       const { data: permissions } = yield call(
         [client, client.listPermissions],
-        { pages: Infinity }
+        { pages: Infinity, filters: { exclude_resource_name: "record" } }
       );
       buckets = expandBucketsCollections(buckets, permissions);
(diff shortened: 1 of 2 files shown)


glasserc:
It took me a while to make sure this wouldn't break editing permissions on records, but actually whenever we edit an object, we issue a batch request (in `getBatchLoadFn`) to fetch the most up-to-date information about it and all its parents.

PR #549, opened 2018-06-07
Remember last used authN method (#525)

### LINKED ISSUE #525: Remember authentication method
- Login successfully to a server with basicauth
- Work
- Logout
- On the auth form, the previous server is filled but the selected authentication method is "anonymous"

leplatrem:
Fixes #525 

- [x] store server and authtype in history
- [x] do not fail with legacy history
- [x] fix flow
- [x] load last used auth type


leplatrem:
@magopian I forgot to tell you about this... 

Would you have some time to pair on it this week?

magopian:
Ok, so this should now be fully functional, and it sadly when as I expected: the "tying into the local state that is also available in the redux store, at the proper place, when we get the server info back" was a pain, and the resulting code isn't very maintainable I'm afraid.

It might make sense in the future to refactor some parts of it to only have one source of truth (the redux store), which might defeat some of the use of react-json-schema-form for this login form. That being said, it might make sense to NOT use rjsf for this form.

I added a test, but it isn't "complete": I'm open to ideas on how to simulate the answer from the `getServerInfo` saga that changes the `capabilities` from the server in the redux store, and thus triggers a `componentDidUpdate` which then sets the `authType` to the value we have in the history.
If we used [enzyme](http://airbnb.io/enzyme/) we'd have a [setState](http://airbnb.io/enzyme/docs/api/ShallowWrapper/setState.html), which would allow us to easily test this kind of corner cases, but I'm not sure it makes sense to add this dependency to this project which is in "maintenance mode" afaik.

@leplatrem @glasserc thoughts?

src/components/AuthForm.js
```diff
@@ -512,6 +513,51 @@ export default class AuthForm extends PureComponent<
...
+      session: {
+        serverInfo: { capabilities: prevCapabilities },
+      },
+    } = prevProps;
+    const {
+      session: {
+        serverInfo: { capabilities: newCapabilities },
+      },
+    } = this.props;
+    if (JSON.stringify(prevCapabilities) !== JSON.stringify(newCapabilities)) {
```
leplatrem: Maybe here we should only compare the list of capabilities keys?

src/components/AuthForm.js
```diff
@@ -512,6 +513,51 @@ export default class AuthForm extends PureComponent<
...
+    } = this.state;
+
+    if (prevServer !== newServer) {
+      // Server changed, set the authType back to "anonymous"
+      return this.onChange({
+        ...this.state,
+        formData: { ...this.state.formData, authType: ANONYMOUS_AUTH },
+      });
+    }
+
```
leplatrem: Please add a comment: `We will look for this server in the history to reuse the auth type...`

src/components/AuthForm.js
```diff
@@ -512,6 +513,51 @@ export default class AuthForm extends PureComponent<
...
+        serverInfo: { capabilities: newCapabilities },
+      },
+    } = this.props;
+    if (JSON.stringify(prevCapabilities) !== JSON.stringify(newCapabilities)) {
+      // The auth methods have changed following a successful "getServerInfo",
+      // update the default auth method with the one from the history, if we
+      // have one for this server.
+      const serverHistoryEntry = this.props.history.find(
+        entry => entry.server === newServer
+      );
```
leplatrem: does it fit on 1 line if written like this ? `({ server }) => server === newServer`

src/components/AuthForm.js
magopian: Actually, I believe we should simply test that the previous capabilities where empty, which is the default value when changing a server (and reinitializing the capabilities)

src/components/AuthForm.js
magopian: Nope, but I can still write it like that 👍

test/components/Homepage_test.js
```diff
@@ -220,14 +220,43 @@ describe("HomePage component", () => {
...
           settings: {},
           session: { authenticated: false, serverInfo: DEFAULT_SERVERINFO },
         });
 
         expect(node.querySelector("#root_server").value).eql(
           "http://server.test/v1"
         );
+        expect(node.querySelector("#root_authType").value).eql("basicauth");
+
+        // Changing the server to another element from the history: the
```
glasserc: Unfinished comment?

src/components/AuthForm.js
magopian: Done that on line 526 and 527, thanks for the heads up

magopian:
I added Enzyme and used it to test setting the `authType` to the one we have in the server history, let me know what you think.

test/components/Homepage_test.js
```diff
@@ -240,23 +243,58 @@ describe("HomePage component", () => {
...
+                  providers: [
+                    {
+                      name: "google",
+                    },
+                  ],
+                },
+              },
+            },
+          },
+        });
```
glasserc: This might be a dumb question, but is setting a giant blob of props the "done thing" in Enzyme? I would have expected some kind of simulated action or something.

magopian: So what Enzyme gives us is the ability to even do `setState` at all (thus changing the state during the test).

Another way then manually setting the props could have been to
1/ use the `HomePage` **container** (and not component)
2/ connect it to a (fake) store
3/ dispatch the action on the store so that it changes the state of the `HomePage` container, and trickles down as props in the `AuthForm`

I feel this would couple code a lot more, and maybe add some indirection (the test wouldn't be as "unit-y"). The compromise is to have the test coupled to the "props/state structure" which is also a problem. Not sure what's the best, what do you think?

glasserc: I don't have a strong opinion, this is fine.

PR #566, opened 2018-07-02
Better auth error message

### LINKED ISSUE #559: Unclear errror message with OpenId Connect
screenshot from 2018-06-28 18-08-27

it shouldn't say «and UserID is undefined». Also it shouldn't show `openid-auth0` since it's our internal sauce :)

magopian:
Fixes #559


src/components/AuthForm.js
```diff
@@ -310,7 +310,7 @@ const authSchemas = authType => {
   };
 };
 
-const getAuthLabel = authType => {
+export const getAuthLabel = (authType: string) => {
```
leplatrem: Maybe put it in `../utils` instead ?

Issue #534, opened 2018-05-18
No package-lock.json


peterbe:
Running `npm install` inside kinto-admin will create a package-lock.json which is not ignored. It also means that the npm packages installed in CI, prior to deciding to make a release are potentially different.

magopian:
Fixed by https://github.com/Kinto/kinto-admin/commit/88ee8b09a20e815ef356c9b27f2754f090253f29

PR #576, opened 2018-07-26
Update outdated dependencies

magopian:
Fixes #512, #555, #534


leplatrem:
Don't we have a lock file?

magopian:
We don't. Yet. cf #534. Also this is not done, I'm now trying to update all the dependencies that have major version changes.

magopian:
This should now be good to merge, not much to review I guess...

package.json
```diff
@@ -36,9 +36,9 @@
     "atob": "^2.0.3",
     "babel-polyfill": "^6.23.0",
     "babel-runtime": "^6.11.6",
-    "bootstrap": "^3.3.6",
+    "bootstrap": "^4.1.3",
```
glasserc: This seems a bit intense to me... don't we have to change markup for this?

magopian: Yeah, you're right, I'll revert that, I'm not going to dig into visual changes, at least for now ;)

PR #585, opened 2018-08-16
Downgrade query-string to 5.1.1

glasserc:
Apparently version 6 is not compatible with create-react-app for
reasons that I don't really understand. See
https://github.com/sindresorhus/query-string/pull/148.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "kinto-admin",
-  "version": "1.19.2",
+  "version": "1.20.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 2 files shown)


peterbe:
Perhaps a more interesting discussion is, why do we need `query-string` at all?

glasserc:
It looks like it's being used to let people link to HistoryTable searches. It's used in a handful of places.

I'm not sure why the CI failed, it looks like the types aren't lined up somehow for reasons that aren't immediately obvious to me.

Also, refs Kinto/kinto#1739.

glasserc:
Also, the reason we didn't notice when we upgraded is that its incompatibility is with a version of react-scripts which we don't use in this project (viz., we don't use any).

peterbe:
> why do we need query-string at all?

What I'm actually referring to is that [`query-string`](https://www.npmjs.com/package/query-string)
is actually a *kinda* a polyfill for [URLSearchParams](https://twitter.com/jaffathecake/status/969551357442183169)

See [this commit](https://github.com/mozilla-services/tecken/commit/1b302919f84f61889fb1824247126deb14f88d32) where I realized that I don't have to worry about [using `URLSearchParams`](https://caniuse.com/#search=URLSearchParams) in old browsers.

Also, check out [this blog post](https://www.peterbe.com/plog/filtertoquerystring) about how I extract the current filter object, from a React state store, into a URL.

glasserc:
Thanks for the suggestion. Closing in favor of #587.

PR #587, opened 2018-08-17
Drop query-string

glasserc:
Move to URLSearchParams instead. This won't work on IE or Edge < 17
but we already disavow any support for those browsers in the
README. Hopefully this will allow better compilation of
kinto-admin (see
https://github.com/sindresorhus/query-string/pull/148).

If this works, it replaces #585.


glasserc:
Thanks @peterbe for the suggestion.

PR #611, opened 2018-09-26
Fix #609: Fix presence of ID value in record edit forms

### LINKED ISSUE #609: If id field has custom title/description, its value is not loaded
screenshot from 2018-09-26 11-16-08

screenshot_2018-09-26 kinto web administration

leplatrem:
fixes #609

### DIFF
--- a/src/components/record/RecordForm.js
+++ b/src/components/record/RecordForm.js
@@ -50,12 +50,12 @@ export function extendUIWithKintoFields(
   isCreate: boolean
 ): Object {
   return {
+    ...uiSchema,
     id: {
       "ui:widget": "text",
       "ui:disabled": !isCreate,
       ...uiSchema.id,
     },
-    ...uiSchema,
   };
 }
 
--- a/src/components/record/RecordForm.js
+++ b/src/components/record/RecordForm.js
@@ -45,6 +45,20 @@ export function extendSchemaWithAttachment(
   };
 }
 
+export function extendUIWithKintoFields(
+  uiSchema: Object,
+  isCreate: boolean
+): Object {
+  return {
+    id: {
+      "ui:widget": "text",
+      "ui:disabled": !isCreate,
+      ...uiSchema.id,
+    },
+    ...uiSchema,
+  };
+}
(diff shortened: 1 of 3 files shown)


leplatrem:
screenshot from 2018-09-26 11-24-10

test/components/RecordAttributes_test.js
```diff
@@ -0,0 +1,98 @@
...
+    it("should show the id as disabled", () => {
+      const field = node.querySelector("#root_id");
+      expect(field.value).to.eql("abc");
+      expect(field.hasAttribute("disabled")).to.be.true;
+    });
+
+    it("should show the id as disabled", () => {
+      const field = node.querySelector("#root_foo");
+      expect(field.value).to.eql("bar");
+    });
```
glasserc: What's this test about? The name doesn't seem to match.

leplatrem:
> what's the intended behavior for the id field?

For some use cases, it makes sense to be able to specify the ID field while creating a record. For some others, the notion of record ID can be completely hidden.

src/components/record/RecordForm.js
```diff
@@ -45,6 +45,20 @@ export function extendSchemaWithAttachment(
...
 
+export function extendUIWithKintoFields(
+  uiSchema: Object,
+  isCreate: boolean
+): Object {
+  return {
+    ...uiSchema,
+    id: {
+      "ui:widget": "text",
+      "ui:disabled": !isCreate,
```
glasserc: Should this maybe be `ui:readonly` instead of disabled?

leplatrem: Good question...

We don't need the `id` value to get passed with the form data.

Should the id field be focusable? included in the tabbed field navigation?

glasserc: I just thought a user might want to copy-paste the ID. I guess they can already grab it from the breadcrumbs, though, so maybe not a big deal?

leplatrem: I would leave it disabled :)

PR #626, opened 2018-09-27
Fix #529, Fix #617, Fix #618: Default Kinto fields

### LINKED ISSUE #529: Hide schema fields by default

### LINKED ISSUE #617: Attachment field should be hidden in the create form and disabled in the edit form
Use `ui:widget: hidden' and 'ui:disabled: true' for the `attachment` field

https://github.com/Kinto/kinto-admin/blob/2ac8c633c4e05b447c1079b43477e58db800d205/src/components/record/RecordForm.js#L47-L64

Otherwise it's very ugly:
screenshot from 2018-09-27 15-56-05

screenshot from 2018-09-27 15-56-38

Whereas it should be: 
screenshot from 2018-09-27 15-57-39

### LINKED ISSUE #618: Attachment.original fields should be hidden if not present
screenshot from 2018-09-27 15-56-38

leplatrem:
* [x] Wait for #611 
* [x] Fix #529 Fix #617  Fix #618
* [x] Add tests

### DIFF
--- a/css/styles.css
+++ b/css/styles.css
@@ -352,7 +352,7 @@ table.record-list td.load-more {
 .attachment-info table {
   margin-bottom: 0;
   flex-grow: 1;
-  display: block; /* for Flex */
+  display: block; /* flex does not support display:table on items. */
 }
 
 .attachment-info .table-condensed > tbody > tr > td {
--- a/css/styles.css
+++ b/css/styles.css
@@ -343,8 +343,16 @@ table.record-list td.load-more {
   padding: 0;
 }
 
+.attachment-info .attachment-attributes {
+  display: flex;
+  align-content: stretch;
+  align-items: end;
+}
+
 .attachment-info table {
   margin-bottom: 0;
+  flex-grow: 1;
+  display: block; /* for Flex */
 }
 
 .attachment-info .table-condensed > tbody > tr > td {
@@ -352,21 +360,16 @@ table.record-list td.load-more {
   text-overflow: ellipsis;
 }
 
-.attachment-action input[type=button] {
-  margin-top: -1em;
-}
-
 .attachment-img {
   display: flex;
   align-items: center;
   justify-content: center;
-  position: absolute;
+  max-width: 150px;
+  max-height: 150px;
   text-align: center;
-  right: 3.2em;
-  width: 125px;
-  height: 125px;
   background: #fff;
   border: 1px solid #ddd;
+  margin: 10px;
 }
 
 .attachment-img img {
--- a/src/components/record/RecordForm.js
+++ b/src/components/record/RecordForm.js
@@ -143,6 +143,10 @@ function AttachmentInfo(props: AttachmentInfoProps) {
     return null;
   }
 
+  const attachmentURL = buildAttachmentUrl(record, capabilities);
+
+  const FileSize = ({bytes}) => filesize(bytes);
+
   return (
     <div className="panel panel-default attachment-info">
       <div className="panel-heading">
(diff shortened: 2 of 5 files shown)


leplatrem:
screenshot from 2018-09-27 20-31-55

screenshot from 2018-09-27 20-31-26
screenshot from 2018-09-27 20-31-38

leplatrem:
I will do something like that (not so ugly):

screenshot from 2018-10-03 18-56-54

leplatrem:
screenshot from 2018-10-04 11-49-16

src/components/record/RecordForm.js
```diff
@@ -119,60 +142,88 @@ function AttachmentInfo(props: AttachmentInfoProps) {
...
+                <th>Hash</th>
+                <td>{attachment.hash}</td>
+              </tr>
+              <tr>
+                <th>Mime-Type</th>
+                <td>{attachment.mimetype}</td>
+              </tr>
+            </tbody>
+          </table>
+          {!!attachment.original && (
```
leplatrem:
```
> kinto-admin@1.20.1 flow-check /home/mathieu/Code/Mozilla/kinto-admin
> flow check

Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/components/record/RecordForm.js:202:44

Cannot get attachment.original.hash because property hash is missing in undefined [1].

     src/components/record/RecordForm.js
     199│                   <td>{filesize(attachment.original.size)}</td>
     200│                 </tr>
     201│                 <tr>
     202│                   <td>{attachment.original.hash}</td>
     203│                 </tr>
     204│                 <tr>
     205│                   <td>{attachment.original.mimetype}</td>

     src/types.js
 [1]  27│   original?: {
      28│     filename: string,
      29│     size: number,
      30│     hash: string,
      31│     mimetype: string,
      32│   },

```
How I can tell Flow that this is safe??

glasserc: Shouldn't this be `attachment.original &&`, not `!attachment.original &&`?

glasserc: Oops, I misread it, it's a double `!`..

test/components/RecordAttributes_test.js
```diff
@@ -127,9 +127,33 @@ describe("RecordAttributes component", () => {
...
+          ...props,
+          capabilities: { attachments: { base_url: "" } },
+          record: gzipped,
+        });
+      });
+
+      it("should show original file attributes", () => {
+        expect(
+          node.querySelector(
+            ".attachment-info table:nth-child(2) tr:nth-child(3) td"
```
glasserc: Is there no class name or anything we can hook onto? This test seems a little brittle.

test/components/RecordAttributes_test.js
```diff
@@ -226,5 +250,214 @@ describe("RecordAttributes component", () => {
...
+          expect(field.value).to.eql("123");
+        });
+
+        it("should show a custom field", () => {
+          expect(field.getAttribute("type")).to.eql("text");
+        });
+
+        it("should show the field as disabled", () => {
+          expect(field.hasAttribute("disabled")).to.be.true;
+        });
```
glasserc: I guess I don't really understand these tests. Should these fields be hidden *only* if they're not in the UI schema? Should we be checking that they are/are not hidden in the case where they are in the UI schema? Similarly, should they be disabled *only* if they're in the UI schema? Should we be checking that they are/are not disabled in the case where they aren't in the UI schema?

css/styles.css
```diff
@@ -343,30 +343,33 @@ table.record-list td.load-more {
...
+.attachment-info .attachment-attributes {
+  display: flex;
+  align-content: stretch;
+  align-items: end;
+}
+
 .attachment-info table {
   margin-bottom: 0;
+  flex-grow: 1;
+  display: block; /* for Flex */
```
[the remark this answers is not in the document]
leplatrem: flexbox does not work with item with `display: table`, I'll adjust the comment

PR #630, opened 2018-09-28
Fix #629: always show default bucket

### LINKED ISSUE #629: Default bucket not shown if server is empty
If the user has not yet created their default bucket, then it's not even shown in the list.

Reported in https://github.com/Kinto/kinto/issues/1733

leplatrem:
Fix #629

### DIFF
--- a/src/sagas/session.js
+++ b/src/sagas/session.js
@@ -225,8 +225,12 @@ export function* listBuckets(
 
     // If the default_bucket plugin is enabled, show the Default bucket first in the list.
     if ("default_bucket" in serverCapabilities) {
-      // Even if server is empty.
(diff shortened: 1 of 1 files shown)


src/sagas/session.js
```diff
@@ -219,6 +222,13 @@ export function* listBuckets(
       }
       data = [];
     }
+
+    // If the default_bucket plugin is enabled, show the Default bucket first in the list.
+    if ("default_bucket" in serverCapabilities) {
+      // Even if server is empty.
+      data = [{ id: userBucket }, ...data.filter(b => b.id != userBucket)];
```
Natim: I guess you should request it first, so that you grab its last_modified value.

PR #639, opened 2018-10-02
Fix #401: Show create bucket only if allowed

### LINKED ISSUE #401: Disable Create bucket if no permission to do it
screenshot from 2017-03-15 16-06-09

Related #86 #135 

> Note: my gnome screenshots have daltonism https://bugs.launchpad.net/ubuntu/+source/gnome-screenshot/+bug/1672716

leplatrem:
Fix #401

### DIFF
--- a/src/components/Sidebar.js
+++ b/src/components/Sidebar.js
@@ -135,6 +135,7 @@ function BucketCollectionsMenu(props) {
 }
 
 type BucketsMenuProps = {
+  canCreateBucket: boolean,
   currentPath: string,
   busy: boolean,
   buckets: BucketEntry[],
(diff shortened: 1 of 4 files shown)


src/components/Sidebar.js
```diff
@@ -170,14 +172,15 @@ function filterBuckets(buckets, filters): BucketEntry[] {
...
+  canCreateBucket: boolean,
   hideReadOnly: boolean,
   search: ?string,
 };
 
 class BucketsMenu extends PureComponent<BucketsMenuProps, BucketsMenuState> {
   constructor(props: BucketsMenuProps) {
     super(props);
-    this.state = { hideReadOnly: false, search: null };
+    this.state = { canCreateBucket: true, hideReadOnly: false, search: null };
```
glasserc: This might be a dumb question but why do we need to put it in the state?

leplatrem: Good catch! I may have done too much while making Flow happy :grin:

PR #642, opened 2018-10-02
[WIP] Fix #641: Fix OpenID auth flow

### LINKED ISSUE #641: OpenID login broken
Nothing happens after redirection.

screencast from 02-10-2018 16_53_23

leplatrem:
Fix #641

### DIFF
--- a/src/components/HomePage.js
+++ b/src/components/HomePage.js
@@ -1,5 +1,10 @@
 /* @flow */
-import type { SessionState, SettingsState, ServerHistoryEntry } from "../types";
+import type {
+  HomePageRouteMatch,
+  SessionState,
+  SettingsState,
+  ServerHistoryEntry,
+} from "../types";
 
 import React, { PureComponent } from "react";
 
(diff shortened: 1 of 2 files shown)


leplatrem:
I could not resist to make it work :)

src/components/HomePage.js
```diff
@@ -69,7 +74,34 @@ type Props = {
   navigateToOpenID: (authFormData: Object, provider: Object) => void,
 };
```
magopian: You need to add the type for `match` here also or else flow will complain (as you can see in the travis build)
```javascript
...
match: HomePageRouteMatch,
...
```

glasserc:
I don't really feel confident in my ability to review this PR. Maybe @peterbe or @rehandalal would be a better choice.

leplatrem:
> I don't really feel confident in my ability to review this PR. Maybe @peterbe or @rehandalal would be a better choice.

I would be very happy if Peter and Rehan could help us redesign the router/redux implementation for the authentication. 

But for this particular case, I just restored some old code that was accidentally dropped.  I will merge it myself once I'm happy with the behaviour

PR #646, opened 2018-10-03
[WIP] Fix #543: Use token type from server

### LINKED ISSUE #543: The auth header type should be taken from the openID Connect capabilities
See https://github.com/Kinto/kinto-admin/pull/509#pullrequestreview-108359050

When authenticating using openID Connect, the headers that are sent to the kinto server are using `Bearer`, which is hardcoded. Instead it should use the `header_type` from the capabilities, eg:

```json
  "openid":{
    "description":"OpenID connect support.",
    "url":"http://kinto.readthedocs.io/en/stable/api/1.x/authentication.html",
    "providers":[
      {
        "name":"google",
        "issuer":"https://accounts.google.com",
        "auth_path":"/openid/google/login",
        "client_id":"248628588820-707f9m6vs19jt7g41cev4iqkglu72u58.apps.googleusercontent.com",
        "header_type":"Bearer",
        "userinfo_endpoint":"https://www.googleapis.com/oauth2/v3/userinfo"
      }
    ]
  }
```

Fixing this might also fix the [review from @leplatrem](https://github.com/Kinto/kinto-admin/pull/509#pullrequestreview-122460014) about the ugliness of splitting/concatenating strings to get the auth type and provider for openID Connect. If we can pass the full "provider capability" object, we'll be able to extract both the provider name and the header type.

leplatrem:
Fix #543

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -8,14 +8,17 @@ let client: ?KintoClient;
 
 function getAuthHeader(auth: AuthData): ?string {
   switch (auth.authType) {
-    case "fxa":
+    case "fxa": {
+      const { token } = auth.credentials;
+      return `Bearer ${token}`;
+    }
     case "openid": {
       const { tokenType, credentials } = auth;
-      const { token }: { token: string } = credentials;
+      const { token } = credentials;
       return `${tokenType} ${token}`;
     }
     case "portier": {
-      const { token }: { token: string } = auth.credentials;
+      const { token } = auth.credentials;
       return "Portier " + token;
     }
     case "anonymous": {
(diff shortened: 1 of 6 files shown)


leplatrem:
> I'm confused. The linked issue mentions header_type, but here it looks like we're using token_type. Also, where does token_type come from? The OpenID server?

You are right. I noticed that the `token_type` was coming in the URL JSON payload and took it from there. It was a lot easier than using our server capabilities metadata. 

But since we have a `header_type` setting on the server, we **should** use it. 
https://kinto.readthedocs.io/en/stable/configuration/settings.html#openid-connect

Thanks for being so thorough, I'll have to re-work this one...

PR #657, opened 2018-10-05
Fix #170: show total number of records

### LINKED ISSUE #170: Display the number of records in a collection
In the collection list of records we should display the total number of records.

leplatrem:
Fix #170


glasserc:
I'm not crazy about this because of Kinto/kinto#1624.

glasserc:
I disagree. I think if we don't merge this, it's still possible to essentially remove `Total-Records` from list-records responses. I think once we merge this, it's much harder to make such a change.

leplatrem:
> I disagree. I think if we don't merge this, it's still possible to essentially remove Total-Records from list-records responses. I think once we merge this, it's much harder to make such a change.

Do you think we should have an explicit call to `collection.getTotalRecords()` instead?

glasserc:
Yes, I would approve of that change.

leplatrem:
screencast from 05-10-2018 15_54_03

I created another action to avoid locking the records loading just for that number

PR #664, opened 2018-10-12
Missing whitespace before link

magopian:
This is a solution taken from https://stackoverflow.com/a/41142436

The issue is visible when creating a new collection:
screen shot 2018-10-12 at 15 28 06

Disclaimer: I haven't tested this.


Natim:
React will never cease to amaze me

leplatrem:
thx :+1:

PR #666, opened 2018-10-15
Fix flow

glasserc:
Fixes #635.


glasserc:
Oops, should have pushed this to #635 instead.

Issue #579, opened 2018-07-27
Display a busy state whenever the server info is fetched


magopian:
At the moment the auth form displays a weird state with only the anonymous auth type available, until the server info and its capabilities are fetched and received. Instead we should
1/ display only the server history bar
2/ as soon as a server is selected (or on mount), when we call `getServerInfo`, display a spinner
3/ when the server capabilities are received, display the proper auth methods radio buttons
4/ if fetching the server capabilities fails, keep the spinner on, display the error message
5/ if another server is selected from the server history dropdown (or a new one entered in the text input), cancel (or discard) the current `getServerInfo` call

This should properly fix issues like #577 or https://github.com/Kinto/kinto-admin/pull/509#issuecomment-389901873

alexcottner:
Closed with [PR 3050](https://github.com/Kinto/kinto-admin/pull/3050).

PR #672, opened 2018-10-23
Fix #615: fix confusing redirect/refresh for auth

### LINKED ISSUE #615: Authentication redirection/reload is ugly/confusing
During the redirection it shows back the login form during a couple of seconds which is kinda confusing.

screencast from 27-09-2018 15_09_56

leplatrem:
Fix #615 

Basically, once authenticated, the server info will never change. So I removed the busy state for the server info, and added an additional state for when the session is being setup. 

The result is a looooooooooooooooot better :) 

### After redirect
screencast from 23-10-2018 16_52_47

### With refresh

screencast from 23-10-2018 16_53_19

### DIFF
--- a/src/components/HomePage.js
+++ b/src/components/HomePage.js
@@ -43,21 +43,17 @@ function ServerProps({ node }: { node: Object }) {
   );
 }
 
-function SessionInfo({ session: { busy, serverInfo } }) {
+function SessionInfo({ session: { serverInfo } }) {
   return (
     <div>
-      {busy ? (
-        <Spinner />
-      ) : (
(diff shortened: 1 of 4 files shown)


leplatrem:
> The screenshots speak for themselves, but I'm a bit confused.. how does the busy state work now in the getServerInfo code path? It seems like there should still be a check for busy somewhere, but I don't see it.

Do you mean when the user enters a URL? There wasn't, see #579

glasserc:
> There wasn't, see #579

Oh. Well then this is clearly an improvement. Merge away.

PR #676, opened 2018-10-24
Fix #624: show a spinner in list during collection metadata load

### LINKED ISSUE #624: Flickering render of records lists (raw JSON to displayFields columns)
screencast from 27-09-2018 18_01_45

leplatrem:
Fix #624


leplatrem:
Alright, so this table will behave a bit differently than the others (collections, groups), because the table columns are dynamic.

PR #692, opened 2018-11-06
Renames "Signed" to "Approved"

### LINKED ISSUE #683: [signoff] Rename "Signed" to "Approved"
screenshot from 2018-10-25 19-55-39

DrFaustie:
Fixes #683

If this is just a change for user experience, this should be adequate. But if it's regarding terminology, there are lots of references to "Signed" within the code such as in sagas.js ` const lastSigned = String(destinationAttributes.last_modified || 0);`

It may be worth updating those too depending on what the intention is

### DIFF
--- a/src/plugins/signoff/components.js
+++ b/src/plugins/signoff/components.js
@@ -150,8 +150,8 @@ export default class SignoffToolBar extends React.Component<
             source={source}
             preview={preview}
           />
-          <Signed
-            label="Signed"
+          <Approved
+            label="Approved"
             step={2}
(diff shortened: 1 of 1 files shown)


src/plugins/signoff/components.js
```diff
@@ -150,8 +150,8 @@ export default class SignoffToolBar extends React.Component<
             source={source}
             preview={preview}
           />
-          <Signed
-            label="Signed"
+          <Approved
```
leplatrem: This should be left as `Signed`

PR #715, opened 2019-01-22
Handle base64 encoded openid token.

Natim:


src/components/HomePage.js
```diff
@@ -93,7 +94,14 @@ export default class HomePage extends PureComponent<Props> {
...
       let tokenType;
       if (authType.startsWith("openid-")) {
-        const parsedToken = JSON.parse(token);
+        let parsedToken;
+        try {
+          // Token is encoded in base64 for a safe path parsing.
+          parsedToken = JSON.parse(atob(token));
+        } catch(e) {
+          // Previous version of Kinto exposed the JSON directly in the URL.
+          parsedToken = JSON.parse(token);
```
Natim: @leplatrem what do you think about this addition?

```suggestion
          try {
            parsedToken = JSON.parse(token);
          } catch (e) {
            console.error(`Token doesn't seems to be a valid JSON: {token}`);
          }
```

PR #929, opened 2019-05-28
Fix #915: Add button to copy authentication header

### LINKED ISSUE #915: Add button to copy Authorization header value to the clipboard
Currently when a user wants to run script, the only way to obtain a Bearer token is to login in the Admin UI, open the dev tools, and copy the autorization header from the completed network requests.

We could at least have a small button close to the username that copies the authorization header value to the clipboard.

leplatrem:
Fix #915

Screenshot from 2019-05-28 13-23-37

/cc @ncloudioj


leplatrem:
Screenshot from 2019-05-28 16-06-57

Don't know how to get out of this with flow. Giving up for today

vanillajonathan:
The button is visible even when using anonymous authentication where there is nothing copy.

leplatrem:
@vanillajonathan Thanks for reporting this! Would you mind opening a new issue or pull-request so that we can track its resolution? Thx

vanillajonathan:
@leplatrem Sure. Created issued #1267.

PR #932, opened 2019-05-28
Fix #501: UI notifications improvements

### LINKED ISSUE #501: Notifications banners should overlay and not move the page content
screenshot from 2018-03-27 17-41-07

This is disturbing that the page content moves...

### LINKED ISSUE #645: Authentication error not shown
When an error is thrown by the OpenID payload/redirection, some other part erase it (during fetch server info I guess)

Using redux devtools I can see that `NOTIFICATIONS_CLEAR` is dispatched:

screenshot from 2018-10-03 12-21-22

leplatrem:
Fixes #501 
Fixes #645

* [x] POC using https://chadly.github.io/react-bs-notifier/
* [x] Fix missing icons (font-awesome vs. glyphicon?)
* [x] Check that errors details are nicely shown
* [x] Fix tests
* [x] Fix flow


leplatrem:
Screencast from 29-05-2019 16_46_55

leplatrem:
I published this branch in the demo: https://kinto.github.io/kinto-admin/

In brief: 
- the notifications disappear automatically (no need to manage persistent / clear)
- the errors have to be closed manually though

Among the things that need to be checked carefully during review:
- is the duration long enough? too long?  
- the notifications during server change / authentication failure etc.
- compare the bundle size (1-1.2Mo)

glasserc:
> the notifications during server change / authentication failure etc.

FYI, I didn't check carefully that the notifications are being handled correctly.

leplatrem:
> I think the code looks good. I think 4 seconds is a little long but I don't feel strongly about it.

Yeah, maybe 3 is enough.

> How does the typing work? I didn't see react-bs-notifier in flow-typed.

Weird. It was not complaining. I added the type file with a dummy file and it failed. So I just put `React.Component`.

>  I think "Authentication is required" is really more like a decoration on the login page than a notification. I don't want to block this PR for that, 

I improved it by adding a success notif. I like the result. Although I wish I could show the `Authentication required.` when it redirects and fails to load the session from local storage. But I couldn't figure out a nice way to do it. So for example, on refresh it stacks 2 notifs «Authentication required.» and «Authenticated - {auth type}»

glasserc:
> I improved it by adding a success notif. I like the result.

We spoke about this in a video chat. This "Authentication required" notification is shown before redirecting to the home page, but the home page can find auth information in local storage, in which case it authenticates you and redirects back to wherever you came from. This means that making the "authentication required" message an error, like I originally suggested, isn't really appropriate, because it could persist even if there's no login form. This might be the best we can do at the moment.

leplatrem:
> This might be the best we can do at the moment.

I've been looking at this, and can't find an elegant way of doing it.

PR #936, opened 2019-05-29
Fix #693 - Drop support of kinto-signer < 3.2

### LINKED ISSUE #693: [signoff] Refactor instances of "Signed" to "Approved"
Follow up to https://github.com/Kinto/kinto-admin/pull/692

It would be best to refactor the current code for consistency and readability

leplatrem:
Fix #693

### DIFF
--- a/src/plugins/signoff/sagas.js
+++ b/src/plugins/signoff/sagas.js
@@ -37,10 +37,12 @@ export function* onCollectionRecordsRequest(getState, action) {
   yield put(SignoffActions.workflowInfo(basicInfos));
 
   // Obtain information for workflow (last update, authors, etc).
-  const {
-    sourceAttributes,
-    changes,
-  } = yield call(fetchWorkflowInfo, source, preview, destination);
+  const { sourceAttributes, changes } = yield call(
+    fetchWorkflowInfo,
+    source,
+    preview,
+    destination
+  );
 
   // Workflow component state.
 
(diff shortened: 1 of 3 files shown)


src/plugins/signoff/sagas.js
```diff
@@ -110,57 +87,21 @@ export function* onCollectionRecordsRequest(getState, action) {
...
-  let lastUpdated;
-  if (sourceChanges.length > 0) {
-    // Some changes exist. Easy, take latest timestamp.
-    lastUpdated = sourceChanges[0].last_modified;
-  } else {
-    // Empty changes list: either never signed, or not changed since last signature.
-    lastUpdated = lastSigned === "0" ? null : sourceAttributes.last_modified;
-  }
+
+  const sourceAttributes = await colClient.getData();
```
glasserc: Kind of a n00b question here, what's the difference between using `yield call` (as we were doing previously) and `await` (as we're doing now)? Was `yield call` an essential part of sagas or was it a workaround because JS didn't have async/await yet?

leplatrem: `yield` is called from generators functions, which are an essential part of sagas. `yield call` is used to call `async` functions or other sagas/generators.

This `fetchWorkflowInfo` function was not a saga, just a util, and can thus be written as a simple async function.

I guess that it was initially written as a generator «by mistake» https://github.com/Kinto/kinto-admin/commit/7db498419739447df3dbb4ce7a68d6403c964bd8#diff-4690b38f7530de693841dfc00502ad16R69

Issue #935, opened 2019-05-29
[signoff] Show diff in the source collection (and not only after review request)


leplatrem:
Otherwise users have no way to see what they've changed until they request a review.

See https://bugzilla.mozilla.org/show_bug.cgi?id=1538967

PR #941, opened 2019-05-30
Add flow to signoff plugin sagas

leplatrem:
Necessary for #935

### DIFF
--- a/src/plugins/signoff/sagas.js
+++ b/src/plugins/signoff/sagas.js
@@ -1,5 +1,12 @@
 /* @flow */
-import type { ActionType, GetStateFn, SagaGen, ServerInfo } from "../../types";
+import type {
+  ActionType,
+  GetStateFn,
+  SagaGen,
+  ServerInfo,
+  BucketResource,
+  CollectionResource,
+} from "../../types";
 import type { ChangesList } from "./types";
 
 import { call, put } from "redux-saga/effects";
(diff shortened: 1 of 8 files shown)


src/plugins/signoff/sagas.js
```diff
@@ -144,19 +190,37 @@ export function* handleDeclineChanges(getState, action) {
     const {
       data: { id: bid },
     } = bucket;
+    if (!bid) {
+      throw new Error("Inconsistent state."); // won't happen (but flow does not know).
+    }
```
glasserc: Maybe we should extract this as a utility?

leplatrem: What would it look like?

glasserc:
```
const bid = assertBucketId(bucket);

function assertBucketId(bucket: BucketResource) {
  const {
    data: { id: bid },
  } = bucket;
  if (!bid) {
    throw new Error("Inconsistent state.");
  }
  return bid;
}
```

BTW, the "true" way to fix this kind of thing is to have two types: one for unsaved buckets (`{last_modified?: number}`), and one for saved buckets (`{id: string, last_modified?: number}`). Sometimes it's worth defining an `Entity` type which wraps the underlying type. I think I considered this at one point but it's a lot harder to retrofit so I sort of let it go..

Issue #520, opened 2018-04-19
[signoff] Rollback changes to latest signed


leplatrem:
> Note: I'm pretty Jorge created this issue somewhere but can't find it :(

PR #943, opened 2019-05-31
Fix #935: show diff in source collection

### LINKED ISSUE #527: [signoff] Simplify and improve consistency when fetching list of changes
See https://github.com/Kinto/kinto-admin/pull/523#issuecomment-384225245

This could be improved:

https://github.com/Kinto/kinto-admin/blob/ae4232f60029e94a75414deee0b519f4fed6ed17/src/plugins/signoff/sagas.js#L114-L121

We should:
- use the destination collection timestamp as `lastSigned`
- use the source collection timestamp as `lastUpdated`

> note: the "never signed" situation should never happen anymore (see https://github.com/Kinto/kinto-signer/pull/227) 

Hint:
```js
const { bucket: bid, collection: cid } = destination;
const { last_modified: lastSigned } = await client.bucket(bid).collection(cid).listRecords();
```

### LINKED ISSUE #935: [signoff] Show diff in the source collection (and not only after review request)

leplatrem:
Fix #935

Fix #527

### DIFF
--- a/src/plugins/signoff/types.js
+++ b/src/plugins/signoff/types.js
@@ -23,7 +23,7 @@ export type SourceInfo = {
   bid: string,
   cid: string,
   // Full info.
-  changes?: ChangesList,
+  changesOnPreview?: ChangesList,
   lastEditBy?: string,
   lastEditDate?: number,
   lastReviewRequestBy?: string,
--- a/src/plugins/signoff/types.js
+++ b/src/plugins/signoff/types.js
@@ -4,6 +4,8 @@ export type CollectionsInfo = {
   source: SourceInfo,
   destination: DestinationInfo,
   preview: ?PreviewInfo,
+  changesOnSource?: ?ChangesList,
+  changesOnPreview?: ?ChangesList,
 };
 
 export type SignoffState = {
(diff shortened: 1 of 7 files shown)


src/plugins/signoff/types.js
```diff
@@ -23,8 +23,8 @@ export type SourceInfo = {
   bid: string,
   cid: string,
   // Full info.
-  changesOnSource?: ChangesList,
-  changesOnPreview?: ChangesList,
+  changesOnSource?: ?ChangesList,
+  changesOnPreview?: ?ChangesList,
```
glasserc: Maybe a comment here explaining that they're absent if preview is disabled and/or only one is present depending on status. I think only one depending on status is reasonable but I think it could use a comment explaining why (maybe "If work-in-progress, show changes since the last review request. Preview cannot be signed directly in this status. Otherwise, show changes that are to be reviewed."). Alternately, is it possible that a user will want to see changes vs. destination as part of a complicated change/review? In that case, maybe we should fetch both.

leplatrem:
> where an editor gets feedback that their settings are bad and they need to completely redo their changes.

Yeah, I want to do #520 at some point

PR #948, opened 2019-06-03
Fix #712: fix history pagination loading

### LINKED ISSUE #712: "Load more" on history pages removes all table content
In history pages:

- clic load more
- the table is emptied
- page scrolls to top
- the results are loaded
- page scroll to bottom

leplatrem:
Fix #712


leplatrem:
> Could you elaborate a little bit on this? Doesn't this mean there's no feedback when loading?

There is already a spinner on the last row of the `PaginatedTable` when loading. 

This was «erasing» the whole table on each page load.

leplatrem:
Oops :rofl:

glasserc:
Thanks! I looked for another `Spinner` but didn't look in `PaginatedTable`. r+

PR #950, opened 2019-06-03
Fix #949: fix history list from signoff diff details

### LINKED ISSUE #949: [signoff] History list in details view contains an extra record
On a signed collection

- clear a record
- click «view details» in the diff info
- the history list should contain one record, and now 2

leplatrem:
Fix #949

### DIFF
--- a/src/components/HistoryTable.js
+++ b/src/components/HistoryTable.js
@@ -14,7 +14,7 @@ import PaginatedTable from "./PaginatedTable";
 import { getClient } from "../client";
 import { omit, sortHistoryEntryPermissions } from "../utils";
 
-function Diff({ source, target }) {
+function Diff({ source, target }: { source: Object, target: Object }) {
   const diff = diffJson(target, source);
   return (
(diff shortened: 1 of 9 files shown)


leplatrem:
> Hmm... are kinto history entries not dated the same as the modifications they represent?

Nope the history entries have their own timestamps. But you're right I should give a detailed example (tomorrow)

leplatrem:
Comment added https://github.com/Kinto/kinto-admin/issues/949#issuecomment-499036218

glasserc:
Thanks for the detailed explanation. I kind of wish that history entries had the same timestamp as the record that had been modified, but maybe that's unreasonable.

leplatrem:
> Thanks for the detailed explanation. I kind of wish that history entries had the same timestamp as the record that had been modified, but maybe that's unreasonable.

Well, they don't. 
But we have an excuse: timestamps are guarenteed to be unique per bucket/collection, whereas history entries are per bucket, which makes it difficult ;)

PR #960, opened 2019-06-05
Fix #443 - Collapse diffs and only show 3 lines of context

### LINKED ISSUE #443: A collapsed mode that display only lines that changed in the diff
It would be nice to have a collapse feature on the diff so that we can directly see the lines that changed.

leplatrem:
Fix #443
Screenshot from 2019-06-05 18-06-55


leplatrem:
> This describes "context", but I think it will also truncate changes too, won't it? 

Good point, we should only truncate the unchanged parts!!! Thanks!

I'll add a few tests...

PR #1517, opened 2020-10-07
Update to the latest version of redux-saga

dstaley:
This PR updates to the latest version of `redux-saga`. One of the first mentioned breaking changes for v1 is that "errors thrown during put execution are no longer caught and swallowed". This means that we're now logging considerably more errors than before, all of which aren't important to tests actually passing. This PR disables `console.error` during a specifically noisy set of tests. In the future we should examine the calls and determine if we really need to be reporting those errors as test failures, but in the meantime this gets us back to where we were prior to upgrading.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3621,6 +3621,53 @@
         }
       }
     },
+    "@redux-saga/core": {
+      "version": "1.1.3",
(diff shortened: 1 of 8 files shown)


leplatrem:
:heart:

PR #1658, opened 2021-01-08
Switch to GitHub actions

leplatrem:


leplatrem:
Weird that `flow` does not find react-router :shrug:

dstaley:
I actually have a branch switching this to TypeScript, so we don't need to worry about flow for much longer. This gets a 👍 from me!

dstaley (approved):

PR #1668, opened 2021-01-18
Convert from Flow to TypeScript

dstaley:
This PR replaces all of the Flow code in the repo with the equivalent TypeScript code. In almost every scenario, it's a one-to-one conversion.

This PR does include a patch for `kinto-http` to export required types. This patch is already in `master` for `kinto-http`, but I need to write up release notes and make a new release. So for now the patch can get things building and passing CI.

The most complex change is to the Redux-related code. It can probably be improved, but I feel pretty confident that it's as good as the existing Flow implementation.

It does not convert the tests to TypeScript, since they weren't Flow-typed anyway.


leplatrem (approved):
Dylan, this is awesome!

Sorry for not getting back to you earlier, it took me a while to dig into this :) 

I could build it locally without issue, let's move forward and merge :tada: :tada:

PR #1701, opened 2021-02-11
Show spinner while loading history

leplatrem:


leplatrem:
Current approach in this PR has a flaw, it looses context while going through pagination

PR #1885, opened 2021-05-25
Simple Review Page

k88hudson:
This is a proposal for a new "simple-review" page. If differs from the main review UI in that:
* Emphasizes the review request comment
* It shows diffs *by record* and compares `main`/`main-workspace` v.s. previous versions

Is this something you would accept? I need some help getting the tests to run properly (they seem to be failing locally for me on `master`)

To test:
1.  With an account with edit privileges, add some changes to a collection and request review (test add, remove, updating records)
2.  With an account with review privileges, visit  `buckets/main-workspace/collections/{aCollection}/simple-review`

image


leplatrem:
Oh! This is awesome! :tada: :tada: Of course we would accept it :) 

I know that @Standard8 was frustrated with the current UI and its multiple confusing screens... Let's iterate together!

Standard8:
Oooh yes, this is a nicer UI.

Per record is similar to what we do in the comparison option of [searchengine-devtools](https://github.com/mozilla-extensions/searchengine-devtools). The diffs for search-config can be quite big, so we use a select box to select the record to look at.

Standard8:
Also, for searchengine-devtools, we show the full diff of the record. The search-config diff records are quire large and repeat various lists of locales etc, so you really do need the full diff - could be an optional switch though.

k88hudson:
I updated with a bunch of tests and rollback button for the `work-in-progress` state, I think it's ready enough if you want to take a look.

Also @Standard8 what do you mean by "full diff"?

k88hudson:
Screenshots of updates:
image
image

k88hudson:
Ahh makes sense, thanks. That should be easy enough to add to add as an optional param yeah

k88hudson:
@leplatrem I think this is ready for you too take a look now

src/components/simpleReview/SimpleReview.tsx
```diff
@@ -0,0 +1,105 @@
...
+  });
+
+  useEffect(() => {
+    async function getRecords() {
+      if (isReviewableBucket && isReviewer && bid && cid) {
+        setRecords({ newRecords: [], oldRecords: [], loading: true });
+        const oldRecords = await fetchRecords(
+          bid.replace("-workspace", ""),
+          cid
+        );
```
leplatrem, line 60: Unfortunately this won't work everywhere.
exposing/reusing `_pickSignoffResource()` in `plugins/signoff/sagas` is the way to go.

```js
const { destination: { bucket: dbid, collection: dcid }} = _pickSignoffResource(serverInfo, bid, cid);
const oldRecords = await fetchRecords(dbid, dcid);
```

src/containers/simpleReview/SimpleReviewPage.ts
```diff
@@ -0,0 +1,37 @@
...
+import { bindActionCreators } from "redux";
+import { connect } from "react-redux";
+
+import SimpleReview from "../../components/simpleReview/SimpleReview";
+import * as SignoffActions from "../../plugins/signoff/actions";
+import { getClient } from "../../client";
+
+function fetchRecords(bid: string, cid: string): Promise<ValidRecord[]> {
+  const coll = getClient().bucket(bid).collection(cid);
+  return coll.listRecords().then(({ data: records }) => {
```
leplatrem, line 14: You can ask the server to sort the records if you prefer: https://github.com/Kinto/kinto-http.js/#sorting

test/components/simpleReview/PerRecordDiffView_test.tsx
```diff
@@ -0,0 +1,124 @@
...
+    expect(node.querySelectorAll(".record-diff")).to.have.lengthOf(1);
+  });
+});
+
+describe("findChangeTypes", () => {
+  it("should find added items", () => {
+    expect(findChangeTypes([{ id: "a" }], [{ id: "a" }, { id: "b" }])).eql([
+      { id: "b", changeType: ChangeType.ADD, target: { id: "b" } },
+    ]);
+  });
```
leplatrem, line 66: nit: add blank line?

leplatrem (commented):
That's awesome! :100: :100: 

Thanks for diving into this ambitious improvement :) 

This is my first feedback without testing it too much on a real server. I'll be back with my feedback on UI/UX :pray:

src/components/simpleReview/SimpleReviewButtons.tsx
```diff
@@ -0,0 +1,62 @@
...
+          <button
+            className="btn btn-danger"
+            onClick={() => setDialog("rollback")}>
+            Rollback
+          </button>
+        )}
+      </div>
+      {dialog === "reject" && (
+        <CommentDialog
+          description="Leave a comment for the editor:"
```
leplatrem, line 46: We could add a sentence here in order to explicit that the changes won't be undone (?)

src/components/simpleReview/PerRecordDiffView.tsx
```diff
@@ -0,0 +1,235 @@
...
+          <div className="form-check form-check-inline mb-3">
+            <input
+              className="form-check-input"
+              type="checkbox"
+              checked={showExtraFields}
+              onChange={e => setShowExtraFields(e.currentTarget.checked)}
+              id="showExtraFields"
+            />
+            <label className="form-check-label" htmlFor="showExtraFields">
+              Show <code>last_modified</code> changes
```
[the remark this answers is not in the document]
k88hudson: Yep I can change that. Re: hide v.s. show, I found the UX a bit confusing when it's in the negative, but me know what you think when you test it out

leplatrem:
I tried it against the stage server, and it's great!

- I liked the workflow of Decline, where Rollback shows up only in work-in-progress state. 
- Show or hide does not really matter. What is important is that the default state matches the basic usage (ie. timestamp hidden)
- The page loading is a bit different than the rest of the app. Adding a `<Spinner/>` may be enough
- I noticed that the actions of approve/decline etc. do not have any spinner and can be a bit confusing when there's latency. But that already applies to the current review UI, so no big deal.
- There is no way to navigate to the simple review mode from the UI. We can update the direct links in the notification emails, but if that's not too much, maybe add a link/switch in the current review UI

Bottom line: it's a huge improvement!

Thanks again! :clap:

leplatrem (approved):

k88hudson:
Ok thanks, I'll get to those review comments and we can move forward on it :) 

> There is no way to navigate to the simple review mode from the UI. We can update the direct links in the notification emails, but if that's not too much, maybe add a link/switch in the current review UI

Might be worth shipping it "dark" and ask a few select people to test it out / give some feedback with real use cases first, after which we could link it / recommend it from the main UI as a follow up.

leplatrem:
> Might be worth shipping it "dark" and ask a few select people to test it out / give some feedback with real use cases first, after which we could link it / recommend it from the main UI as a follow up.

Why not, yes!

PR #2013, opened 2021-08-25
Fix Bootstrap 5

vanillajonathan:
The `form-group` class is removed in Bootstrap 5, the recommended replacement is `mb-3`.

https://getbootstrap.com/docs/5.1/migration/

### DIFF
--- a/src/components/Sidebar.tsx
+++ b/src/components/Sidebar.tsx
@@ -264,7 +264,7 @@ class BucketsMenu extends PureComponent<BucketsMenuProps, BucketsMenuState> {
             <strong>Filters</strong>
           </div>
           <form className="form card-body">
-            <div className="form-group">
+            <div className="mb-3">
               <div className="input-group">
                 <input
                   type="text"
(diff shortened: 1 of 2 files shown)


leplatrem:
Thanks!

Should we leave both?

vanillajonathan:
> Should we leave both?

I don't understand what you mean.

https://getbootstrap.com/docs/5.1/migration/#forms

> Dropped form-specific layout classes for our grid system. Use our grid and utilities instead of .form-group, .form-row, or .form-inline.

Their new examples uses the `mb-3` class in place of the previously used `form-group` class.

leplatrem:
I'm sorry, I completely forgot that https://github.com/Kinto/kinto-admin/pull/1857 was merged and that we were running bs5 already :man_facepalming:  The PR was merged from a script without testing it!

This is ok then :) Thanks for the fix!

vanillajonathan:
You might want to bump Bootstrap 5.0 to 5.1.0.

Issue #2135, opened 2021-10-28
Migrate Redux setup to Redux Toolkit


grahamalama:
As [was discussed](https://github.com/Kinto/kinto-admin/issues/2119#issuecomment-952800382) in #2119:
> "It would be great if we could align [our Redux setup] with the latest recommendations."

The latest recommendations are essentially to use Redux Toolkit (RTK). Much of the Redux documentation today assumes you are using Redux Toolkit.

My thought for migration would be:
1. migrate slices of the store (one slice at a time) to [createSlice](https://redux-toolkit.js.org/api/createSlice) 
1. modify our store creation setup to use RTK's [configureStore](https://redux-toolkit.js.org/api/configureStore)
1. update documentation to instruct plugin authors to migrate to RTK (optional?)

PR #2136, opened 2021-10-28
Add redux toolkit

grahamalama:
Opening this PR as a draft to demo what a migration to RTK might look like when it comes to how action creators, reducers, and sagas will change. (#2135)

The sessions slice was pretty straightforward, but I could see the slices that use `paginator()` being a little less easy, though I haven't really dived in that much.

If this is something we want to tackle, I can mark it as ready for review. If we want to hold off on the migration though, I can delete this PR until later.


grahamalama:
Closing this PR for now, but might consider this migration in the future.

PR #2180, opened 2021-11-16
Downgrade to history v4

leplatrem:
See https://github.com/supasate/connected-react-router/issues/312

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,7 +5,6 @@
   "requires": true,
   "packages": {
     "": {
-      "name": "kinto-admin",
       "version": "1.31.0",
       "hasInstallScript": true,
       "license": "Apache-2.0",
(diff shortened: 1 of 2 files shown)


grahamalama (commented):
Can you think of a way we could write a test for this that fails with `history@^5` but passes with `history@^4` to prevent future regressions?

leplatrem:
> Can you think of a way we could write a test for this that fails with history@^5 but passes with history@^4 to prevent future regressions?

Something around following links / navigation. I don't know yet whether this would require a full browser or not.

grahamalama:
Alternatively, maybe we can upgrade our build process to use npm v7 because of how [peer dependencies](https://github.blog/2021-02-02-npm-7-is-now-generally-available/#peer-dependencies) are treated:
> In previous versions of npm (4-6), peer dependencies conflicts presented a warning that versions were not compatible, but would still install dependencies without an error. npm 7 will block installations if an upstream dependency conflict is present that cannot be automatically resolved.

leplatrem:
Good catch! Yes! I was indeed wondering why peer deps were not blocking anything here

grahamalama (approved):

PR #2182, opened 2021-11-16
Revert "Bump react, react-dom from 16.14.0 to 17.0.2, use new test adapter (#2055)"

grahamalama:
This reverts commit cbd7e6a35116488fa0e3e4341dcd0e30f77013e4.

Reverting because `react-breadcrumbs` [specifies React as a peer dependency](https://github.com/robbestad/react-breadcrumbs/blob/40598df6d152ec01b9a720f71c597815be208000/package.json#L35) for only versions `^15.0.0 || ^16.0.0`


leplatrem (approved):
As you mentioned in #2180, using a recent Node version would have helped us catch this

PR #2184, opened 2021-11-16
Bump Node version to used in Github actions to 16

grahamalama:
Bumps `Node.js` to `^16`, [the LTS version](https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V16.md#16.13.0) at the time of this PR. This also bumps `npm` to version 8.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,8 +28,8 @@
   ],
   "engineStrict": false,
   "engines": {
-    "npm": "^2.14.7",
-    "node": ">=10"
+    "npm": ">=8",
+    "node": ">=16"
   },
   "dependencies": {
     "@babel/polyfill": "^7.0.0",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -26,10 +26,9 @@
     "images",
     "lib"
   ],
-  "engineStrict": false,
   "engines": {
-    "npm": "^2.14.7",
-    "node": ">=10"
+    "npm": ">=8",
+    "node": ">=16"
```
leplatrem, line 31: Not sure whether we should require it here, but that's ok

leplatrem (approved):
Thanks :pray:

matt-boris (approved):
Nice! 🎉

PR #2205, opened 2021-11-30
Add a Selenium browser test

leplatrem:


grahamalama:
@leplatrem are we good to close this since we're going to be maintaining integration tests in `mozilla/remote-settings`? Or is this something we want to maintain here as well?

leplatrem:
> @leplatrem are we good to close this since we're going to be maintaining integration tests in mozilla/remote-settings? Or is this something we want to maintain here as well?

As mentionned elsewhere, having these browser tests in this repo would allow to make sure we detect issues as soon as they occur (ie. pull-request), instead of downstream in Kinto plugin or Remote Settings.

One idea could be to try out https://playwright.dev

leplatrem:
👋

PR #2207, opened 2021-11-30
Replace react-breadcrumbs with custom component

### LINKED ISSUE #2181: Find alterrnative to react-breadcrumbs (abandonned?)
The repo wasn't updated since 2019.

And it has React 16 in peer dependencies...

The code isn't too big, maybe we could vendor it.

grahamalama:
Closes #2181

[React Breadcrumbs](https://github.com/robbestad/react-breadcrumbs) was causing us some headaches since it specifies React version `15.x` or `16.x` as a peer dependency. Its last commit was made in 2019, so it appears to be abandoned.

We weren't really using much from this package and it was easy enough to roll our own breadcrumbs component, so this PR does just that.

While working on this PR, I noticed that `history` was bumped to version `5.x` (which is incompatible with `connected-react-router`), so this PR also includes a downgrade of history to major version 4.

Before:
Screen Shot 2021-12-01 at 9 18 18 AM

After:
Screen Shot 2021-12-01 at 9 21 10 AM


src/utils.tsx
```diff
@@ -409,3 +410,20 @@ export function diffJson(
...
 }
+/**
+ * Given a pathname, return a 2d array where each inner array is the path
+ * part's name and location
+ *
+ * "/"        => [["home", "/"]]
+ * "/foo"     => [["home", "/"], ["foo", "/foo"]]
+ * "/foo/bar" => [["home", "/"], ["foo", "/foo"], ["bar", "/foo/bar"]]
+ */
+export function breadcrumbifyPath(path: string) {
```
matt-boris, line 421: Nice!

matt-boris (approved):
🎉

leplatrem (approved):
:100: great job

PR #2305, opened 2022-01-28
Combine actions into one CI workflow

grahamalama:
- Take advantage of matrixes to test and build on windows and ubuntu
- Cache node_modules
- Remove kinto install since we don't run integration tests in any of
  these workflows


leplatrem (approved):
Thanks for this :tada:

PR #2372, opened 2022-03-17
Tweak webpack for HTML inclusion in bundle and setting Public Path

grahamalama:
This PR:

Adds [HtmlWebpackPlugin](https://webpack.js.org/plugins/html-webpack-plugin/) to handle including an HTML file in the bundled app. With this plugin:
- html is minified
- the publicPath option is set correctly
- there's no need to manually handle sending html from the dev server
- includes a favicon 🖼️

It also makes `publicPath` configurable via an `ASSET_PATH` environment variable ([example from docs](https://webpack.js.org/guides/public-path/#environment-based)). We set `ASSET_PATH=/kinto-admin/` when building for Github Pages.

Finally, it deletes the `server/test.js` server. I couldn't find usage of it anywhere.


webpack.prod.js
```diff
@@ -1,8 +1,11 @@
 var path = require("path");
 var webpack = require("webpack");
 const MiniCssExtractPlugin = require("mini-css-extract-plugin");
+const HtmlWebpackPlugin = require("html-webpack-plugin");
 var version = require("./package.json").version;
 
+const ASSET_PATH = process.env.ASSET_PATH || "/";
```
leplatrem, line 7: Maybe here we can add a comment explaining that this will be set to this and that in gh-pages, kinto plugin, etc.

leplatrem (approved):
Nice!

webpack.prod.js
grahamalama: https://github.com/Kinto/kinto-admin/pull/2372/commits/9068653c117810b6d09f11754b11160a868064d8

PR #2373, opened 2022-03-19
Use environment variable to set `SINGLE_SERVER`, deprecate other settings

grahamalama:
In this PR, we:
- move `sidebarMaxListedCollections` and `maxPerPage` to plain constants that live in `src/constants.ts` since we didn't override the defaults in the Kinto Admin plugin
- read `SINGLE_SERVER` as an environment variable (rather than a prop that gets passed to the `KintoAdmin` component)
- remove the `settings` slice of the redux store from throughout the app. Now, we import and use constants when needed


src/components/App.tsx
```diff
@@ -111,8 +111,7 @@ export default class App extends PureComponent<Props> {
   render() {
     const { copyAuthenticationHeader, session, logout } = this.props;
     const contentClasses = `col-sm-9 content`;
-    const version =
-      process.env.REACT_APP_VERSION || process.env.KINTO_ADMIN_VERSION;
```
leplatrem, line 115: This would deserve an entry in the CHANGELOG, as a breaking change?

src/utils.tsx
```diff
@@ -291,19 +291,13 @@ export function debounce(fn: any, delay: number) {
...
   // Return the server URL value, by priority:
   // - single server mode
   // - most recently used
   // - default
-  return (
-    singleServer ||
-    (servers && servers.length && servers[0].server) ||
-    DEFAULT_KINTO_SERVER
-  );
+  console.table({ SINGLE_SERVER });
```
leplatrem, line 299: Should we log `servers` here too? (last used servers)

webpack.prod.js
```diff
@@ -27,8 +27,11 @@ module.exports = {
     }),
     new MiniCssExtractPlugin({ filename: "styles.css" }),
     new webpack.DefinePlugin({
-      "process.env.KINTO_ADMIN_VERSION": JSON.stringify(version),
-      "process.env.ASSET_PATH": JSON.stringify(ASSET_PATH),
+      "process.env": {
+        ASSET_PATH: JSON.stringify(ASSET_PATH),
```
leplatrem, line 31: This maybe deservers a mention in the README?

leplatrem (approved):
Nice :ok_hand:

webpack.prod.js
grahamalama: I think it does too. I was going to take care of that in a follow-up PR because I think a good section of the documentation will have to change if we're going to encourage downloading a release a building from source rather than `npm install` ing the `KintoAdmin` component.

src/utils.tsx
grahamalama: 🤦🏻 that was accidentally left in from `console` debugging. Unless you think that logging would actually be helpful, I was thinking I'd remove this rather than add more.

src/components/App.tsx
grahamalama: This would be something to include in the release notes of the next release, right? Since we don't have a CHANGELOG in this repo (yet).

leplatrem: Ah! I thought you added it ;)

PR #2377, opened 2022-03-22
Remove published <KintoAdmin /> component, tweak Webpack config

grahamalama:
This PR removes references to building / publishing Kinto Admin to NPM as a standalone `<KintoAdmin />` component.

It also slightly rearranges the top-level components of the app to remove any traces of the `<KintoAdmin />` component:

- the old `App.tsx` component becomes the `Layout` component since that's where we lay out the header, sidebar, main content, and footer
- new `App.tsx` for Redux, Router, and `Layout`
- `src/index.tsx` for main entrypoint (where we call `ReactDOM.render()`)
- remove `index.js` at top level of project

It also slightly tweaks our Webpack setup to use the [Webpack 5 way of loading assets](https://webpack.js.org/guides/asset-modules/#root).


leplatrem (approved):
Tested locally! Looks sound :clap:

PR #2378, opened 2022-03-23
Migrate away from container components, use `react-redux` hooks

grahamalama:
In this PR, we start the process of removing "container" components from throughout the app where we `map{State, Disptach}ToProps`. Instead, we'll use [react-redux hooks](https://react-redux.js.org/api/hooks) to provide dispatch and Redux store values to components. This will lead to an overall fewer number of components to manage, which is a plus.

This PR adds a typed useSelector hook -- a typed useDispatch hook will have to come later because the way we set up the initial store is a little atypical.

We also started the process with Layout and Sidebar because they were relatively simple components to migrate. 

(I'll probably stick to migrating 2 or 3 components at a time going forward.)


leplatrem (approved):
Nice! 
Tested locally, and works nicely

PR #2391, opened 2022-03-30
Modify `createComponent` test helper to accept JSX

grahamalama:
This PR updates our `createComponent` helper to accept a component with props as JSX (`createComponent(<Component prop={prop} />, ...)` rather than `createComponent(Component, {props}, ...)`. 

This subtle change will make the work of migrating container components easier. It also aligns us more with the [API of `testing-library`](https://testing-library.com/docs/react-testing-library/setup#custom-render), which might be something we want to think about using for our component tests.


leplatrem (approved):
LGTM

PR #2403, opened 2022-04-12
Remove Permissions containers

grahamalama:
Remove the `{Group, Bucket, Collection, Record}Permissions` container pages in favor of redux / react-router hooks. 

And instead of passing `bid` and `groups` params to the permissions form component, use them directly in the component with redux hooks.


leplatrem (approved):
Testing locally and seems to work fine :)

PR #2426, opened 2022-04-28
Refactor redux setup

grahamalama:
This PR:
#### Adds [Redux Toolkit](https://redux-toolkit.js.org/) for Redux store setup
We're not going all-in with Redux toolkit patterns yet, but this PR does use it for its `combineReducers` and new `configureStore` API to set up middleware (the core `redux` `createStore` method was recently [marked as deprecated](https://github.com/reduxjs/redux/issues/4325)).

#### Converts `<App />` to a functional component
In a continuing effort to convert class based components to functional components, we converted the App component in this PR. The tricky part with this component was that it used an unconventional method of setting up the redux store in the class constructor method. This made it convenient since in the same place we set up the Redux store we could also fire an initial saga that set up the session, but when we moved to a functional component, we no longer had access to the store or dispatch through hooks.

Since the initial setup only affects the presentation of the `HomePage` component though, we were able to move the that initial setup logic there. Tests also moved the the `HomePage_test` file.


package.json
```diff
@@ -57,7 +57,7 @@
     "@babel/preset-react": "^7.16.7",
     "@babel/preset-typescript": "^7.16.7",
     "@svgr/webpack": "^6.0.0",
-    "@types/history": "^5.0.0",
+    "@types/history": "^4.7.11",
```
grahamalama, line 60: I think Dependabot keeps bumping this dependency, but we want it back at major version 4 to align with our version of history that we need to keep for `connected-react-router`.

src/components/HomePage.tsx
```diff
@@ -155,41 +180,27 @@ export default class HomePage extends PureComponent<Props> {
       const message = "Couldn't proceed with authentication.";
       notifyError(message, error);
     }
-  };
+  }, []);
```
grahamalama, line 183: This empty dependency array (and the one above) essentially makes this useEffect equivalent to `componentDidMount`.

src/hooks.ts
```diff
@@ -1,4 +1,5 @@
-import { TypedUseSelectorHook, useSelector } from "react-redux";
-import type { AppState } from "./types";
+import { TypedUseSelectorHook, useSelector, useDispatch } from "react-redux";
+import type { AppDispatch, AppState } from "./types";
 
 export const useAppSelector: TypedUseSelectorHook<AppState> = useSelector;
+export const useAppDispatch = () => useDispatch<AppDispatch>();
```
grahamalama, line 5: https://redux-toolkit.js.org/usage/usage-with-typescript#getting-the-dispatch-type

src/store/configureStore.ts
```diff
@@ -1,39 +1,28 @@
...
-      ? ((window as any).__REDUX_DEVTOOLS_EXTENSION__() as () => void)
-      : (f) => f
-  )(createStore);
-
-  const store = finalCreateStore(
-    createRootReducer(history),
-    initialState as any
-  );
+export function configureAppStore(initialState = {}, history = hashHistory) {
+  const store = configureStore({
```
grahamalama, line 13: See https://redux-toolkit.js.org/api/configureStore#basic-example

src/store/configureStore.ts
```diff
@@ -1,39 +1,28 @@
...
+    reducer: createRootReducer(history),
+    middleware: getDefaultMiddleware =>
+      getDefaultMiddleware({
+        thunk: false,
+      }).concat(sagaMiddleware, routerMiddleware(history)),
```
grahamalama, line 18: The `defaultMiddleware` includes some nice additions including devtools without manual setup and warnings when save something to the store that isn't serializable.

leplatrem (approved):
I read the changes, and diff per commit, they all seem to make sense. Thanks for the clean history, that helps 🙏 

If you expect a deep review, I would need to learn a bit more about the redux toolkit, which I haven't really thoroughly before reviewing

PR #2443, opened 2022-05-12
Remove Homepage container

grahamalama:
This PR continues the work of replacing the "container component" pattern with connecting components to the Redux store using hooks.

This PR also slightly restructures `Homepage` tests. In addition to changing tests to account for the new method of connecting the component to the store, we also moved some tests from `Homepage` to `Authform`, since those tests were really more suited for that test suite.


src/components/HomePage.tsx
```diff
@@ -193,14 +160,18 @@ export default function HomePage(props: Props) {
...
-          clearServers={clearServers}
-          navigateToExternalAuth={navigateToExternalAuth}
-          navigateToOpenID={navigateToOpenID}
+          clearServers={() => dispatch(ServersActions.clearServers())}
+          navigateToExternalAuth={authFormData =>
+            dispatch(SessionActions.navigateToExternalAuth(authFormData))
+          }
+          navigateToOpenID={(authFormData, provider) =>
+            dispatch(SessionActions.navigateToOpenID(authFormData, provider))
+          }
```
grahamalama, line 174: This is pretty ugly, but temporary. Eventually, we'll be able to remove many of these props and access this redux state / dispatch actions directly from `AuthForm`.

leplatrem (approved):
Thanks for the step by step commits ;)

PR #2498, opened 2022-06-21
Use https

vanillajonathan:


leplatrem (approved):
Good catch! Thank you :)

vanillajonathan:
The link to the kinto admin demo website in the "About" section is still on http though, and it is also broken, it returns 404.

leplatrem (approved):

PR #2504, opened 2022-06-27
Replace `connected-react-router` with `redux-first-history`

grahamalama:
Development seems to have slowed somewhat in `connected-react-router`, and as such, it has us pinned us at old major versions of several important libraries because of the [peer dependencies](https://github.com/supasate/connected-react-router/blob/master/package.json#L34:L40) that they specify:
| Library | Peer dependency version | Current version |
|--------------|-----------|------------|
history | `^4.7.2` | `5.3.0`
react | `^16.4.0 \|\| ^17.0.0` | `18.2.0`
react-redux | `^6.0.0 \|\| ^7.1.0` | `8.0.2`
react-router | `^4.3.1 \|\| ^5.0.0` | `6.3.0`

When looking at issues in that repo, I came across another library ([`redux-first-history`](https://github.com/salvoravida/redux-first-history/)) that connects the router to the Redux store and specifies [fewer peer dependencies](https://github.com/salvoravida/redux-first-history/blob/master/package.json#L44-L47).

This PR migrates `connected-react-router` to `redux-first-history`. 

I will also mention that as a future issue, we should consider whether we need to store location state in Redux at all. From the "original", now deprecated Redux + React Router library:

> This library is not necessary for using Redux together with React Router. You can use the two together just fine without any additional libraries. It is useful if you care about recording, persisting, and replaying user actions, using time travel. If you don't care about these features, just [use Redux and React Router directly](http://stackoverflow.com/questions/36722584/how-to-sync-redux-state-and-url-hash-tag-params/36749963#36749963).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,7 +17,6 @@
         "bootstrap-icons": "^1.0.0",
         "btoa": "^1.1.2",
         "codemirror": "^5.39.2",
-        "connected-react-router": "^6.9.2",
         "diff": "^5.0.0",
         "express": "^4.14.0",
         "filesize": "^8.0.2",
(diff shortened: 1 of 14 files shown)


leplatrem (approved):
LGTM and browser history navigation seems to work smoothly when testing locally 👌

Issue #2159, opened 2021-11-02
Use official Enzyme test adapter when it is available for React 17


grahamalama:
In order to bump React to the latest version in #2209, we had to install an [unofficial Enzyme adapter](https://github.com/wojtekmaj/enzyme-adapter-react-17) that supported React 17. When Enzyme releases a test adapter that supports React 17 (tracked [here](https://github.com/enzymejs/enzyme/issues/2429)), we should uninstall the unofficial package and use the official Enzyme package.

grahamalama:
Closing this issue for now since we rolled React back to v16 in #2182

alexcottner:
PR has been merged. Enzyme has been removed. Marking issue as closed, although we have more work to follow up on.

PR #3039, opened 2023-10-23
Removing enzyme, adding testing-library/react

alexcottner:
Replacing Enzyme with testing library.

**Note:** this replaces enzyme with testing-library, but there is still more work to do to make our tests better. Created issue https://github.com/Kinto/kinto-admin/issues/3041 to follow up on making our unit tests better.

Issue: #2159


grahamalama:
Nice first pass! Happy to see green tests. 

After a quick look over, I noticed that some of the ways we're testing seem like they're `testing-library` anti-patterns. I also noticed some things where I'm wondering if they're out of scope for this PR:

Where we do `render(...).container`:

> If you find yourself using container to query for rendered elements then you should reconsider! The other queries are designed to be more resilient to changes that will be made to the component you're testing. Avoid using container to query for elements! ([src](https://testing-library.com/docs/react-testing-library/api/#container-1))

Where we use `node.querySelector`:

> React Testing Library aims to test the components how users use them. Users see buttons, headings, forms and other elements by their role, not by their id, class, or element tag name. Therefore, when you use React Testing Library you should avoid accessing the DOM with the document.querySelector API. (You can use it in your tests, but it's not recommended for the reasons stated in this paragraph.) ([src](
https://testing-library.com/docs/react-testing-library/migrate-from-enzyme/#test-2-input-texts-must-have-correct-value))

Where we `import { expect } from "chai";`

Should we use this opportunity to drop Chai? I was able to do this with some success:

<details>
<p>

```diff
diff --git a/test/reducers/group_test.js b/test/reducers/group_test.js
index c7672a0..fbd3fcb 100644
--- a/test/reducers/group_test.js
+++ b/test/reducers/group_test.js
@@ -1,5 +1,3 @@
-import { expect } from "chai";
-
 import group from "../../src/reducers/group";
 import {
   GROUP_BUSY,
@@ -14,8 +12,7 @@ describe("group reducer", () => {
   describe("GROUP_BUSY", () => {
     it("should update the busy flag", () => {
       expect(group(undefined, { type: GROUP_BUSY, busy: true }))
-        .to.have.property("busy")
-        .eql(true);
+        .toHaveProperty("busy", true)
     });
   });
 
@@ -23,15 +20,14 @@ describe("group reducer", () => {
     it("should reset group state", () => {
       const initial = group(undefined, { type: null });
       const altered = group(initial, { type: GROUP_BUSY, busy: true });
-      expect(group(altered, { type: GROUP_RESET })).eql(initial);
+      expect(group(altered, { type: GROUP_RESET })).toEqual(initial)
     });
   });
 
   describe("ROUTE_LOAD_REQUEST", () => {
     it("should set the busy flag", () => {
       expect(group({ busy: false }, { type: ROUTE_LOAD_REQUEST }))
-        .to.have.property("busy")
-        .eql(true);
+        .toHaveProperty("busy", true)
     });
   });
 
@@ -39,7 +35,7 @@ describe("group reducer", () => {
     it("should preserve state when no group is passed", () => {
       const initial = group(undefined, { type: null });
 
-      expect(group(undefined, { type: ROUTE_LOAD_SUCCESS })).eql(initial);
+      expect(group(undefined, { type: ROUTE_LOAD_SUCCESS })).toEqual(initial);
     });
 
     it("should update state when a group is passed", () => {
@@ -51,16 +47,15 @@ describe("group reducer", () => {
         },
       });
 
-      expect(state.data).eql({ id: "grp", foo: "bar", last_modified: 42 });
-      expect(state.permissions).eql({ read: ["a"], write: ["b"] });
+      expect(state.data).toEqual({ id: "grp", foo: "bar", last_modified: 42 });
+      expect(state.permissions).toEqual({ read: ["a"], write: ["b"] });
     });
   });
 
   describe("ROUTE_LOAD_FAILURE", () => {
     it("should clear the busy flag", () => {
       expect(group({ busy: true }, { type: ROUTE_LOAD_FAILURE }))
-        .to.have.property("busy")
-        .eql(false);
+        .toHaveProperty("busy", false)
     });
   });
 
@@ -84,10 +79,10 @@ describe("group reducer", () => {
 
       const state = group(initial, action);
 
-      expect(state.history.entries).eql([1, 2, 3]);
-      expect(state.history.loaded).eql(true);
-      expect(state.history.hasNextPage).eql(true);
-      expect(state.history.next).eql(fakeNext);
+      expect(state.history.entries).toEqual([1, 2, 3]);
+      expect(state.history.loaded).toEqual(true);
+      expect(state.history.hasNextPage).toEqual(true);
+      expect(state.history.next).toEqual(fakeNext);
     });
   });
 });
diff --git a/test/setup-tests.js b/test/setup-tests.js
index b1f05f3..a180b61 100644
--- a/test/setup-tests.js
+++ b/test/setup-tests.js
@@ -1,3 +1,5 @@
+import '@testing-library/jest-dom'
+
 // stub out scrollTo for tests
 global.window.scrollTo = () => {};
 

```

</p>
</details> 

Similar to above, I wonder if we can use `testing-library`'s `fireEvent` or even [`user-event`](https://testing-library.com/docs/user-event/intro/) so don't need to use `Simulate` from `react-dom/test-utils`


For each of these things, if there's a reason we're "breaking the rules" or you have some follow up work planned to keep PRs small, maybe we can include that in the PR description.

alexcottner:
We actually have a lot of room for improvement here. I can expand this PR and implement some better tests, I was taking the shortest path to replace enzyme as a first pass. But as you mentioned:

1. We should be using test-id's or something instead of query selectors.
2. We should replace other rendering with testing-library rendering
3. We can eliminate Chai (and probably sinon)
4. We can replace user interactions with fireEvent (doing this in one place on testing-library already)

Should I start working on these changes on this PR or after?

grahamalama:
> Should I start working on these changes on this PR or after?

I think as long as we make clear (though the PR title + description) what's in/out of scope for this PR, and for the things that are out of scope (what we've identified here), spin them out out into new issues, I'm happy with this PR being just about `"yeet Enzyme"` 👍🏻.

test/components/AuthForm_test.js
```diff
@@ -241,57 +241,31 @@ describe("AuthForm component", () => {
...
-      // authType is now set to the value we have from the servers history.
-      // TODO: for some reason, we don't have a "input#root_authType" anymore
-      // at this point, so we need to get at the `authType` value some other
-      // way, here by looking into the state.
-      const state = wrapper.find(AuthForm).instance().state;
-      expect(state.formData.authType).eql("openid-google");
+      // TODO: Simulate a `getServerInfo` response that updated the `capabilities`.
+      // The enzyme test has been removed as part of the upgrade to testing-library,
+      // but not replaced because it interracted with properties and state directly.
+      // This needs to be rewritten to mock user input and server response instead.
```
grahamalama, line 268: Have you looked into how much work it'd take to write this test in a `testing-library`-friendly way? It'd be nice to keep this test coverage rather than delete the test.

alexcottner: Attempted to add a test that captures the spirit of the old test. Let me know your thoughts on this.

alexcottner:
Created issue #3401 to follow up on making our unit tests better.

test/components/AuthForm_test.js
grahamalama: Lgtm! A pragmatic workaround.

grahamalama (approved):

PR #3049, opened 2023-11-02
Add spinners and disable forms while data is being loaded

alexcottner:
The goal of this PR is to prevent users from interacting with forms while we're waiting on server responses. I think the easiest way to do that is to:

1. Disable the forms while we're waiting for responses after user submissions/onChange events that trigger API calls
2. Provide a visual cue to users that data is being loaded

This should resolve the following open issues:
1. [Display a busy state whenever the server info is fetched](https://github.com/Kinto/kinto-admin/issues/579)
2. [Prevent double submission of forms](https://github.com/Kinto/kinto-admin/issues/1360)
3. [Simple review should disable the approve/reject buttons after they're clicked](https://github.com/Kinto/kinto-admin/issues/2206)
4. [Loading indicator not shown when loading history](https://github.com/Kinto/kinto-admin/issues/1596)

[Video of this in action](https://github.com/alexcottner/kinto-admin/assets/148472676/d6030b1d-9d4d-4489-8249-caa6682ab3f7)

Issue #2206, opened 2021-11-30
[signoff] Simple Review should disable the approve/reject buttons after they're clicked


jaredlockhart:
Currently, when a user clicks approve or reject, the button remains enabled while the API call is being made with no visual indication and the buttons are still enabled so there's nothing preventing someone from clicking them multiple times.  We should disable the buttons while the API call is being made to prevent users from firing multiple requests.

leplatrem:
See also https://github.com/Kinto/kinto-admin/issues/1360

alexcottner:
Closed with [PR 3050](https://github.com/Kinto/kinto-admin/pull/3050).

Issue #1596, opened 2020-11-09
Loading indicator not shown when loading history


leplatrem:

alexcottner:
Should be fixed with my next MR. See video below.

[Video link here](https://github.com/Kinto/kinto-admin/assets/148472676/52a9d2c1-5fe2-4b76-b941-c1291362e143)

alexcottner:
Closed with [PR 3050](https://github.com/Kinto/kinto-admin/pull/3050).

Issue #1360, opened 2020-05-14
Prevent double submission of forms


leplatrem:
When submitting a record creation form, I accidentally submitted the form twice, and two records were created. 

This should never happen. Submission should be disabled if an action is already being handled.

alexcottner:
Closed with [PR 3050](https://github.com/Kinto/kinto-admin/pull/3050).

PR #3050, opened 2023-11-03
Add spinners and disable forms while data is being loaded - alternate

alexcottner:
Alternate (hopefully slightly cleaner) PR compared to [3049](https://github.com/Kinto/kinto-admin/pull/3049).

The goal of this PR is to prevent users from interacting with forms while we're waiting on server responses. I think the easiest way to do that is to:

1. Disable the forms while we're waiting for responses after user submissions/onChange events that trigger API calls
2. Provide a visual cue to users that data is being loaded

This should resolve the following open issues:
1. [Display a busy state whenever the server info is fetched](https://github.com/Kinto/kinto-admin/issues/579)
2. [Prevent double submission of forms](https://github.com/Kinto/kinto-admin/issues/1360)
3. [Simple review should disable the approve/reject buttons after they're clicked](https://github.com/Kinto/kinto-admin/issues/2206)
4. [Loading indicator not shown when loading history](https://github.com/Kinto/kinto-admin/issues/1596)

[Video of this in action](https://github.com/alexcottner/kinto-admin/assets/148472676/d6030b1d-9d4d-4489-8249-caa6682ab3f7)

### DIFF
--- a/test/components/BaseForm_test.js
+++ b/test/components/BaseForm_test.js
@@ -43,7 +43,7 @@ describe("BaseForm component", () => {
 
   afterEach(() => {});
 
-  it("Should render a jrsf form as expected", async () => {
+  it("Should render a rjsf form as expected", async () => {
     const result = render(
       <BaseForm
         className="testClass"
--- a/test/components/BaseForm_test.js
+++ b/test/components/BaseForm_test.js
@@ -1,0 +1,107 @@
+import React from "react";
+import { render } from "@testing-library/react";
+import BaseForm from "../../src/components/BaseForm";
+
+const testSchema = {
+  type: "object",
+  properties: {
+    title: {
+      type: "string",
+      title: "Title",
+      description: "Short title",
+    },
+    content: {
+      type: "string",
+      title: "Content",
+      description: "Provide details...",
+    },
+  },
+};
+
+const testUiSchema = {
+  "ui:order": ["title", "content"],
+  content: {
+    "ui:widget": "textarea",
+  },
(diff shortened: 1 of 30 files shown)


test/components/BaseForm_test.js
```diff
@@ -0,0 +1,134 @@
...
+    "ui:widget": "hidden",
+    "ui:disabled": true,
+  },
+  "ui:disabled": false,
+};
+
+describe("BaseForm component", () => {
+  beforeEach(() => {});
+
+  afterEach(() => {});
```
grahamalama, line 44: Do we need these?

test/components/BaseForm_test.js
```diff
@@ -0,0 +1,134 @@
...
+
+  it("Should render a jrsf form as expected", async () => {
```
grahamalama, line 46:
```suggestion
  it("Should render a rjsf form as expected", async () => {
```

grahamalama (approved):
Nice! I'll throw an approval here for now, though you said you're still going to add some tests to cover, https://github.com/Kinto/kinto-admin/issues/1360, https://github.com/Kinto/kinto-admin/issues/2206, and https://github.com/Kinto/kinto-admin/issues/1596 right?

test/components/BaseForm_test.js
alexcottner: Nope! Added them out of habit and forgot to remove them because prettier made them too small. I noticed them in the next branch and will be removing them then.

Issue #3045, opened 2023-11-02
Unable to create a collection when logged in as a user, permission error


alexcottner:
Getting an invalid permission error when trying to create a collection (only when not logged in as admin). This appears to be an issue from a recent PR.

![image](https://github.com/Kinto/kinto-admin/assets/148472676/15b9c9c5-e89e-4447-81cf-ec24f69990b0)

alexcottner:
Closed with [PR 3051](https://github.com/Kinto/kinto-admin/pull/3051)

PR #3051, opened 2023-11-03
Bugfixes for collection create/edit and record loading

alexcottner:
Small PR that corrects two bugs and adds unit tests around them. This PR:
1. Corrects a known issue introduced in the functional component conversion [where non-admin users couldn't create a new collection](https://github.com/Kinto/kinto-admin/issues/3045).
2. Corrects an issue I discovered while working on this where collection records didn't load correctly when switching collections/buckets.

### DIFF
--- a/src/components/collection/CollectionForm.tsx
+++ b/src/components/collection/CollectionForm.tsx
@@ -17,6 +17,7 @@ import { canCreateCollection, canEditCollection } from "../../permission";
 import { validateSchema, validateUiSchema } from "../../utils";
 import DeleteForm from "./DeleteForm";
 import { FormInstructions } from "./FormInstructions";
+import { Link } from "react-router-dom";
 
 const defaultSchema = JSON.stringify(
   {
@@ -310,7 +311,7 @@ export default function CollectionForm({
         {` ${creation ? "Create" : "Update"} collection`}
       </button>
       {" or "}
-      <a href="/#/">Cancel</a>
+      <Link to="/">Cancel</Link>
       {" | "}
       <a href="#" onClick={toggleJSON}>
         {asJSON ? "Edit form" : "Edit raw JSON"}
--- a/src/components/collection/CollectionForm.tsx
+++ b/src/components/collection/CollectionForm.tsx
@@ -6,7 +6,6 @@ import type {
 } from "../../types";
 
 import React, { useState } from "react";
-import { Link } from "react-router-dom";
 
 import { Check2 } from "react-bootstrap-icons";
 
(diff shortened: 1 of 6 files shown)


src/components/collection/CollectionForm.tsx
```diff
@@ -311,7 +310,7 @@ export default function CollectionForm({
         {` ${creation ? "Create" : "Update"} collection`}
       </button>
       {" or "}
-      <Link to="/">Cancel</Link>
+      <a href="/#/">Cancel</a>
```
leplatrem, line 313: Out of curiosity, why this change?

test/components/CollectionForm_test.js
```diff
@@ -0,0 +1,142 @@
...
+      width: 0,
+    });
+    Range.prototype.getClientRects = () => ({
+      item: () => null,
+      length: 0,
+      [Symbol.iterator]: jest.fn(),
+    });
+  });
+
+  it("Should render an eitable form for a user with permissions creating a new collection", async () => {
```
leplatrem, line 67: typo: editable

leplatrem, line 67: Nitpick: maybe set the permissions here in the test setup like in other tests with `localTestProps.session.permissions.push()` to improve readability and test atomicity (leaving global `testProps` with an empty array in permissions for example

test/test_utils.js
```diff
@@ -81,3 +82,50 @@ export function sessionFactory(props) {
     redirectURL: "",
   };
 }
+
+// temporary solution until we rewrite AdminLink to not require router around it
```
leplatrem, line 86: Do we have an issue to link here for reference?

leplatrem (approved):
👌 
only questions/nitpicks

src/components/collection/CollectionForm.tsx
alexcottner: Originally it was because we don't need a Link there to do stuff and it was messing with my unit tests. But now I have a better idea that I'm going to implement.

test/components/CollectionForm_test.js
alexcottner: Woops

test/test_utils.js
alexcottner: I've changed my mindset on this one a bit. Mocking Link to avoid issues with react-router is easy and I'm probably going to do that in focused unit tests going forward, so no need to remove it. 

However because we're very tied to redux right now we can't get away from Provider as easily, so there is less of a point on larger unit tests. So I created the `renderWithProvider` function that we can start migrating towards. I'm hoping this causes less friction in short and long term changes with our unit tests in this repo.

test/components/CollectionForm_test.js
alexcottner: That's a good call.

leplatrem (approved):

Issue #2361, opened 2022-03-10
[signoff] List of records is not refreshed after rollback 


leplatrem:
Steps to reproduce: 

- Use an empty collection
- Create a record, status is work in progress
- Click *Rollback*
- Status is back to signed, but record is still present and tab says *Records (1)* 
- Click on the Records tab to reload the page, list of records is empty

Expected:
- list of records should be refreshed (ie. empty) after rollback

alexcottner:
Resolved with last PR

Issue #1953, opened 2021-07-09
[simple-review] Add query param to hide rollback button


k88hudson:
For experimenter usage, we do not want people to trigger a rollback via the UI.

It would be nice to be able to hide the button with a query param (to be put in links in Experimenter UI)

See https://github.com/mozilla/experimenter/blob/6e08fa198b07324b412543246390c0dab046cd7c/experimenter/experimenter/experiments/models.py#L679-L686

alexcottner:
Resolved with last PR

PR #3053, opened 2023-11-08
Record refresh bugfix and simple-review enhancement to hide rollback button

alexcottner:
Resolves two issues and adds unit tests around them:

1. [Add query param to hide rollback button](https://github.com/Kinto/kinto-admin/issues/1953)
2. [List of records is not refreshed after rollback](https://github.com/Kinto/kinto-admin/issues/2361)

**Note:** The added the query param is only useful while the user does not navigate away and back. Should this be a change that needs to be persisted throughout the session please leave a comment and I'll adjust the code to save this to sessionStorage and check against that.

Video demos showing fixes:

[Issue 1953](https://github.com/Kinto/kinto-admin/assets/148472676/e71b7b70-cb58-47bd-a15a-3b25e70e3e06)

[Issue 2361](https://github.com/Kinto/kinto-admin/assets/148472676/f40a6ef1-d4d5-40eb-b48a-374bb2423fd4)

### DIFF
--- a/src/actions/collection.ts
+++ b/src/actions/collection.ts
@@ -57,18 +57,21 @@ export function listNextRecords(): {
 export function listRecordsSuccess(
   records: RecordData[],
   hasNextRecords: boolean,
-  listNextRecords: Function | null | undefined
+  listNextRecords: Function | null | undefined,
+  isNextPage: boolean = false
 ): {
   type: "COLLECTION_RECORDS_SUCCESS";
   records: RecordData[];
   hasNextRecords: boolean;
   listNextRecords: Function | null | undefined;
+  isNextPage: boolean;
 } {
   return {
     type: COLLECTION_RECORDS_SUCCESS,
     records,
     hasNextRecords,
     listNextRecords,
+    isNextPage,
   };
 }
 
--- a/src/locationUtils.ts
+++ b/src/locationUtils.ts
@@ -18,7 +18,8 @@ export function breadcrumbifyPath(path: string) {
   return [["home", "/"]].concat(zip(crumbNames, crumbPaths));
 }
 
-// This may become obsolete when upgrading or switching router packages
+// This may become obsolete when upgrading or switching router packages. Functionality
+// is not built in to react-router v5, useSearchParams exists in v6 though.
 export function parseSearchString(search?: string): Record<string, any> {
   if (!search || !search.startsWith("?")) {
     return {};
--- a/src/locationUtils.ts
+++ b/src/locationUtils.ts
@@ -1,0 +1,35 @@
+import zip from "lodash/zip";
+
+/**
+ * Given a pathname, return a 2d array where each inner array is the path
+ * part's name and location
+ *
+ * "/"        => [["home", "/"]]
+ * "/foo"     => [["home", "/"], ["foo", "/foo"]]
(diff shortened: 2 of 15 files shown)


src/locationUtils.ts
```diff
@@ -0,0 +1,35 @@
...
+  const crumbNames = path.split("/").filter(Boolean);
+  const pathParts = [""];
+  const crumbPaths = crumbNames.map(name => {
+    pathParts.push(name);
+    return pathParts.join("/");
+  });
+  return [["home", "/"]].concat(zip(crumbNames, crumbPaths));
+}
+
+// This may become obsolete when upgrading or switching router packages
```
leplatrem, line 21: nitpick: can we add notes about current/expected versions or upgrade PR number

src/sagas/collection.ts
```diff
@@ -53,7 +53,7 @@ export function* listRecords(
       limit: MAX_PER_PAGE,
     });
     // Show list of records.
-    yield put(actions.listRecordsSuccess(data, hasNextPage, next));
+    yield put(actions.listRecordsSuccess(data, hasNextPage, next, false));
```
leplatrem, line 56: nitpick: can we add a little annotation for this `false`? like:
```suggestion
    yield put(actions.listRecordsSuccess(data, hasNextPage, next, /* isNextPage: */ false));
```

src/sagas/collection.ts
```diff
@@ -71,7 +71,7 @@ export function* listNextRecords(getState: GetStateFn): SagaGen {
   }
   try {
     const { data, hasNextPage, next } = yield call(listNextRecords);
-    yield put(actions.listRecordsSuccess(data, hasNextPage, next));
+    yield put(actions.listRecordsSuccess(data, hasNextPage, next, true));
```
leplatrem, line 74: ditto

src/reducers/collection.ts
```diff
@@ -82,10 +82,10 @@ export function collection(
       return { ...state, recordsLoaded: false };
     }
     case COLLECTION_RECORDS_SUCCESS: {
-      const { records, hasNextRecords, listNextRecords } = action;
+      const { records, hasNextRecords, listNextRecords, isNextPage } = action;
       return {
         ...state,
-        records: [...state.records, ...records],
+        records: isNextPage ? [...state.records, ...records] : records,
```
leplatrem, line 88: I'm not sure that I understand this. Maybe `isNextPage` does not represent what we are really doing here?
Could you please add a comment why we skip `state.records` when `isNextPage`?

src/actions/collection.ts
```diff
@@ -57,18 +57,21 @@ export function listNextRecords(): {
 export function listRecordsSuccess(
   records: RecordData[],
   hasNextRecords: boolean,
-  listNextRecords: Function | null | undefined
+  listNextRecords: Function | null | undefined,
+  isNextPage: boolean = false
```
leplatrem, line 61: What alternative names would we have for this `isNextPage`?

leplatrem (approved):
Great 👌 

Since this PR covers two commits, please don't squash them  🙏  (also, nitpick we usually use `Fix #123: ` or `(Fixes #123)` to reference issues in commit)

Issue #616, opened 2018-09-27
No attachment widget is shown if collection has no schema


leplatrem:
* Create a collection without schema, only attachment: enabled + required

![screenshot from 2018-09-27 15-50-08](https://user-images.githubusercontent.com/546692/46150541-15564800-c26d-11e8-8773-b05f2b41e248.png)

* Record form shows raw JSON without attachment form widget

![screenshot from 2018-09-27 15-51-59](https://user-images.githubusercontent.com/546692/46150653-4d5d8b00-c26d-11e8-8fc5-69bcec0763ed.png)

> Note: we may end up with that at Mozilla via the YAML manifest in the cloudops repo

PR #3056, opened 2023-11-13
Fixes 616 - Adds file attachment logic to JSONRecordForm and tests

alexcottner:
Fixes [issue 616](https://github.com/Kinto/kinto-admin/issues/616) by adding a file attachment input to the JSONRecordForm component.

image

### DIFF
--- a/src/components/record/JSONRecordForm.tsx
+++ b/src/components/record/JSONRecordForm.tsx
@@ -9,37 +9,95 @@ const schema: RJSFSchema = {
   default: "{}",
 };
 
+const schemaWithUpload: RJSFSchema = {
+  type: "object",
+  properties: {
+    jsonContent: {
+      type: "string",
+      title: "JSON record",
+      default: "{}",
+    },
+    __attachment__: {
+      type: "string",
+      format: "data-url",
+      title: "File attachment",
+    },
+  },
+  required: ["jsonContent"],
+};
+
+const uiSchemaWithUpload: UiSchema = {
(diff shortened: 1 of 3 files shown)


src/components/record/RecordForm.tsx
```diff
@@ -170,6 +170,9 @@ export default function RecordForm(props: Props) {
             disabled={!allowEditing}
             record={JSON.stringify(recordData, null, 2)}
             onSubmit={handleOnSubmit}
+            attachmentEnabled={attachment?.enabled}
+            // require an attachment on initial insert if attachments are required
```
leplatrem, line 174: This comment could mention the fact that we don't want to require attachments on existing records, because we want to be able to edit the fields without having to reupload its attachment? Or is it obvious?

src/components/record/JSONRecordForm.tsx
```diff
@@ -9,37 +9,95 @@ const schema: RJSFSchema = {
...
   onSubmit: (data: Object) => void;
   children?: React.ReactNode;
 };
 
+function splitAttachment(record: string) {
+  const jsRecord = JSON.parse(record);
+  const attachment = jsRecord.attachment || {};
+  if (jsRecord.attachment) {
+    delete jsRecord.attachment;
+  }
```
leplatrem, line 58: nitpick: We have `omit()` in utils for this

leplatrem (approved):
👌

Issue #633, opened 2018-09-28
If collection creation is not allowed, form shows confusing message


leplatrem:
- bucket is readonly
- no permission to create collection
- button is shown anyway (see #86)
- Create collection form says something about «edit»
![screenshot from 2018-09-28 16-09-40](https://user-images.githubusercontent.com/546692/46213555-1f457d00-c339-11e8-921d-c074296739c6.png)

alexcottner:
I cannot reproduce this fully, perhaps it's already been fixed? When I see a read-only collection now the "Create collection" button is missing from the sidebar. However I can see it on the collections list. I'll fix the second button showing up.
![image](https://github.com/Kinto/kinto-admin/assets/148472676/f20f10d0-30c7-47a4-9943-1fde9513f32a)

leplatrem:
Indeed, I think both things need to be fixed:
- remove the `Create collection` when readonly
- fix the message `You don't have the required permission to edit this collection.` when a user reaches to the URL directly with `#/buckets/{bid}/collections/create`  (or redirect to list with a notification, if you find that clearer)

PR #3057, opened 2023-11-14
Fixes issue 633 - Hide buttons users shouldn't be able to click

alexcottner:
This PR fixes [issue 633](https://github.com/Kinto/kinto-admin/issues/633).

1. Does not render a "Create collection" button of the user lacks permissions to do create a collection on this bucket
2. Does not render a "Create group" button if the user lacks permissions to edit this bucket
3. Renders a better error if the user happens to land on the Create collection page with a direct link.

With permission:
image

Without permission:
image

Better error message:
image

### DIFF
--- a/src/components/bucket/GroupDataList.tsx
+++ b/src/components/bucket/GroupDataList.tsx
@@ -5,6 +5,7 @@ import { ClockHistory } from "react-bootstrap-icons";
 
 import { timeago } from "../../utils";
 import AdminLink from "../AdminLink";
+import { canEditBucket } from "../../permission";
 
 export function DataList(props) {
   const { bid, groups, capabilities } = props;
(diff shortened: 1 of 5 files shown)


src/components/collection/CollectionForm.tsx
```diff
@@ -296,7 +296,11 @@ export default function CollectionForm({
   const alert =
     allowEditing || bucket.busy || collection.busy ? null : (
       <div className="alert alert-warning">
-        You don't have the required permission to edit this collection.
+        You don't have the required permission to{" "}
```
alexcottner, line 299: why prettier, why did you do this to me

src/components/bucket/GroupDataList.tsx
```diff
@@ -66,7 +67,7 @@ export function DataList(props) {
 }
 
 export function ListActions({ bid, session, bucket }) {
-  if (session.busy || bucket.busy) {
+  if (session.busy || bucket.busy || !canEditBucket(session, bucket)) {
```
leplatrem, line 70: This should probably be `!canCreateGroup()` no?

alexcottner: Wow I totally missed that function and thought it was lumped in with bucket edit. Thank you for catching that.

leplatrem (approved):

Issue #621, opened 2018-09-27
Cannot edit collection metadata without display fields defined


leplatrem:
* Create a collection without display fields
* Edit attributes
* Cannot save. 

--> it should be left undefined or left with id and last_modified as default

alexcottner:
@leplatrem - Is this issue no longer valid? It looks like we throw a custom validation error telling users we require at least one display field. 
![Screenshot from 2023-11-14 10-00-39](https://github.com/Kinto/kinto-admin/assets/148472676/192d69c2-16c1-4d2c-865a-f300a7babc49)

leplatrem:
Sorry @alexcottner I must have been clearer on the description.

The first step has to be done via the API. For example:

```
http PUT "${SERVER}/v1/buckets/main-workspace/collections/test-empty-metadata" "Authorization: ${AUTHZ}"
```

alexcottner:
Oh perfect, that was my missing context. So in that case we have a mismatch between our UI validation and what the API allows in. I'm guessing the API is correct here so I will open these rules up to allow for inserts/updates without display fields.

PR #3058, opened 2023-11-14
Fixes 621 - Allow users to create/edit collections without displayFields

alexcottner:
Fixes [issue 621](https://github.com/Kinto/kinto-admin/issues/621) by:
1. Allowing users to create and edit collections without displayFields objects defined (reduced minimum array size and updated validation message)
2. Fixed collection records default display columns to work when an empty array or a nullish object are passed in

image

image

### DIFF
--- a/src/components/collection/CollectionForm.tsx
+++ b/src/components/collection/CollectionForm.tsx
@@ -81,7 +81,7 @@ const schema: RJSFSchema = {
       title: "Records list columns",
       description: "fields to list in the record list table",
       default: ["title"],
-      minItems: 1,
+      minItems: 0,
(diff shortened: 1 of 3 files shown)


leplatrem (approved):
💯

leplatrem:
Could you please use `#` in the PR titles / commits?

Usually we set the message to: `Fix #123: msg` or `Msg (fixes #123)`

alexcottner:
> Could you please use `#` in the PR titles / commits?
> 
> Usually we set the message to: `Fix #123: msg` or `Msg (fixes #123)`

Got it, will remember that going forward.

Issue #3059, opened 2023-11-15
Attempting to update a collection's settings when there is an error, leads to no obvious indication without scrolling


alexcottner:
Copied from [bugzilla bug 1593229](https://bugzilla.mozilla.org/show_bug.cgi?id=1593229).

> STR
>
>    Edit an existing collection's setup, by clicking the cog icon.
>    On a short-ish screen (so that the update button is off the page).
>    Do something to create an error, e.g. insert a comma to make a schema invalid.
>    Scroll down and click the "Update Collection" button
>
>Expected Results
>
>=> Indication there's an error, and what the issue is.
>
>Actual Results
>
>=> Nothing obvious happens.
>=> Only when you scroll back up to the top of the page do you see the errors listed.

This appears to affect everything using our `BaseForm` component. Main plan is to create generic error handling there (assuming rjsf doesn't have something already) so we scroll to the top-most validation error when we fail validation.

PR #3060, opened 2023-11-15
Fixes issue #3059 - scroll to first validation error on BaseForm

alexcottner:
Fixes [issue 3059](https://github.com/Kinto/kinto-admin/issues/3059). When a user receives a validation error for a specific property, it will now scroll and center the label for that property in their view. 

If an error is thrown for the form rather than a specific property, it will scroll to the top of the form where the full errors list appears.

[demo video](https://github.com/Kinto/kinto-admin/assets/148472676/6902e40d-6221-457e-8318-9944e3cdedc0)

### DIFF
--- a/test/components/BaseForm_test.js
+++ b/test/components/BaseForm_test.js
@@ -148,6 +148,7 @@ describe("BaseForm component", () => {
     titleLabel.scrollIntoView = testFn;
     fireEvent.click(submit);
     expect(testFn).toHaveBeenCalledTimes(1);
+    expect(result.container.querySelector(`[for="root_title"]`)).not.toBeNull();
   });
 
   it("Should scroll to the top of the form if validation failed without a specific property", async () => {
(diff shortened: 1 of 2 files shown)


test/components/BaseForm_test.js
```diff
@@ -127,4 +127,48 @@ describe("BaseForm component", () => {
...
+        className="testClass"
+        schema={testSchema}
+        uiSchema={testUiSchema}
+        onSubmit={jest.fn()}
+        customValidate={(data, errors) => {
+          errors.title.addError("test error");
+          return errors;
+        }}
+      />
+    );
```
leplatrem, line 143: Shall we add a test here to make sure the query selector that we use in the code is found in the page? `.querySelector(`[for="root${err.property.replace(/\./g, "_")}"]`)`
This way if RJSF changes the way it names its fields, the test would break

alexcottner: Good call, added

leplatrem (approved):

PR #3076, opened 2023-11-29
Fixes #3073 - don't get stuck when collection data cannot load for anonymous users

### LINKED ISSUE #3073: Trying to view certain collections with `anonymous` auth leads to endless loading spinners across the app
Steps to reproduce:

- log into https://remote-settings-dev.allizom.org/v1/admin/#/ with anonymous `auth`
- Check `Show readonly buckets/collections`
- Click on `addons` collection in the `blocklist` bucket
- Result:
image
In the console:
image

- Try to navigate back home and click the "refresh" button for collections:
image

alexcottner:
This is the simplest solution I could come up with for https://github.com/Kinto/kinto-admin/issues/3073 . This is not the full fix, however I think this makes release v2.1.1 stable.

Please review https://github.com/Kinto/kinto-admin/pull/3074 before reviewing this one.

A test site has been published to [gh pages](https://alexcottner.github.io/kinto-admin).


src/sagas/signoff.ts
```diff
@@ -45,106 +45,116 @@ export function* onCollectionRecordsRequest(
...
-  yield put(SignoffActions.workflowInfo(collectionsInfoWithChanges));
+    const collectionsInfoWithChanges = {
+      ...collectionsInfo,
+      changesOnSource,
+      changesOnPreview,
+    };
+    yield put(SignoffActions.workflowInfo(collectionsInfoWithChanges));
+  } catch (ex) {
+    if (ex.data?.code === 401) {
+      console.warn(ex);
```
leplatrem, line 153: Can't we just do this around the call to `yield call(fetchSourceAttributes, source);` ?

src/components/signoff/SimpleReview/index.tsx
```diff
@@ -102,9 +102,20 @@ export default function SimpleReview({
...
+          const oldRecords = await fetchRecords(destBid, destCid);
+          setRecords({ oldRecords, newRecords, loading: false });
+        } catch (ex) {
+          if (ex.data?.code === 401) {
+            setRecords({
+              ...records,
+              loading: false,
+            });
+          } else {
+            console.error(ex);
```
leplatrem, line 116: Can we show an error notification?

leplatrem (approved):

src/sagas/signoff.ts
alexcottner: Nope, unfortunately. This was fun to troubleshoot, but there were multiple places in this saga that can end up giving us a 401 response. (ex: SignoffActions.workflowInfo can as well, and that's called from several places).

I figured one big try/catch was better than 10 small ones since the end result was all the same.

src/components/signoff/SimpleReview/index.tsx
alexcottner: That's a good idea, added notification and updated tests.

PR #3077, opened 2023-11-29
Fixes #2506 - Updating react codemirror package that we use

alexcottner:
Fixes https://github.com/Kinto/kinto-admin/pull/2506. Be sure to review https://github.com/Kinto/kinto-admin/pull/3076 first as this is branched from that.

Switching to a react codemirror package that supports codemirror v6. Differences are subtle but present. Would like to hear any objections over using this package, but it seems solid to me.

**New**
Screenshot from 2023-11-29 13-20-45

**Old**
Screenshot from 2023-11-29 13-21-31


src/App.tsx
```diff
@@ -4,7 +4,7 @@ import { Provider } from "react-redux";
 
 import { store, history } from "./store/configureStore";
 import "bootstrap/dist/css/bootstrap.css";
-import "codemirror/lib/codemirror.css";
+// import "codemirror/lib/codemirror.css";
```
leplatrem, line 7: not necessary anymore? Shall we just remove it?

src/components/JSONEditor.tsx
```diff
@@ -39,9 +24,11 @@ export default function JSONEditor({
   ) : (
     <CodeMirror
       value={value}
-      autoCursor={false}
       onChange={onCodeChange}
-      options={cmOptions}
+      height="auto"
+      style={{}}
+      extensions={[langs.json()]}
+      indentWithTab={true}
```
leplatrem, line 31: We had `indentWithTabs: false,` with options like `tabSize: 2,`. Why this change?

src/App.tsx
alexcottner: Yup, missed that one.

leplatrem (approved):

PR #3078, opened 2023-11-30
Fixes #687 - moving sessionStorage to localStorage

### LINKED ISSUE #687: When I ctrl-click on the edit button, I need to authenticate again
While trying to find a way to edit multiple records in batch, I've tried to open the edition form in as-many separate tabs as records I wanted to edit.

But when going to these newly created tabs, I'm presented with the authentication form (even if I'm already identified !)

alexcottner:
Fixes https://github.com/Kinto/kinto-admin/issues/687 by moving sessionStorage to localStorage. Now users can open multiple tabs without having to login again. Auth will also persist between tab/window closes.

Please review https://github.com/Kinto/kinto-admin/pull/3077 as this is branched from that.

Note: this fixes the immediate problem. Longer term I'd like to migrate this to the `useLocalStorage` hook. But we can't do that from the current redux sagas.


alexcottner:
> Since we don't use refresh tokens, I wonder if this is fully safe to store bearer forever... BTW do our bearer tokens expire? And how does the UI react when they do?
When the token expires, or can't be validated, the user gets redirected to the login form as you would expect.

Our tokens are JWE's and not JWS's, so I can't inspect them to find an exact expiration date. Investigating further.

If needed I can switch the auth storage to be in a cookie. Unfortunately it has to only be auth info as the session object is too big to keep in a cookie.

alexcottner:
The tokens last 24 hours. Would you rather manually set a shorter timeout via a cookie?

grahamalama (approved):

Issue #989, opened 2019-06-27
Login method is not restored


leplatrem:
Steps to reproduce
1. Open the Admin UI
2. Select a login method
3. Login successfully
4. Revisit the Admin UI
5. The login method should be preselected

![Screenshot from 2019-06-27 15-14-52](https://user-images.githubusercontent.com/546692/60269253-64113b00-98ee-11e9-914e-6342f2943b28.png)

PR #3080, opened 2023-11-30
Issue 989 - Fixing login type in server history

alexcottner:
This fixes https://github.com/Kinto/kinto-admin/issues/989 by storing the full auth type in server history rather than the shortened type. LDAP auth methods should be remembered correctly now.


leplatrem (approved):
🎉

PR #3094, opened 2023-12-11
Resolves #3041 - Finishing some test cleanup

### LINKED ISSUE #3041: Update UI unit tests to be better
Update our unit tests to be better. We've replaced enzyme with testing-library already, so this should be a _relativley_ easy lift.

1. Use test-id's (or similar) instead of query selectors.
2. Replace other rendering with testing-library rendering
3. Eliminate chai and probably sinon to simplify our dependencies
4. Replace user interactions with fireEvent (or similar)
5. Remove all state interaction/checks in tests. UI tests should be done against elements rendered.

alexcottner:
This PR finishes resolving https://github.com/Kinto/kinto-admin/issues/3041 by removing querySelectors from tests (at least for the parent elements being grabbed) and removing form submit calls and replaced them with button clicks.

This PR also removes some tests that were deep testing dependency components instead of our code. I don't think we should be testing things this deeply as we don't have control over what's happening upstream in bootstrap or rjsf.

Component changes were limited to adding test id's and a label `for` attribute.

### DIFF
--- a/src/components/HomePage.tsx
+++ b/src/components/HomePage.tsx
@@ -22,7 +22,7 @@ function ServerProps({ node }: { node: any }) {
         {nodes.map(([key, childNode], i) => {
           return (
             <tr key={i}>
-              <th>{key}</th>
+              <th data-testid="home-th">{key}</th>
               <td style={{ width: "100%" }}>
                 {isObject(childNode) || Array.isArray(childNode) ? (
                   <ServerProps node={childNode} />
@@ -47,7 +47,7 @@ function SessionInfo({ session: { serverInfo } }) {
   return (
     <div>
       <div className="card server-info-panel">
-        <div className="card-header">
+        <div className="card-header" data-testid="home-header">
           <b>Server information</b>
         </div>
         <div className="card-body">
--- a/src/components/HomePage.tsx
+++ b/src/components/HomePage.tsx
@@ -47,7 +47,7 @@ function SessionInfo({ session: { serverInfo } }) {
   return (
     <div>
       <div className="card server-info-panel">
-        <div className="card-header" data-testid="home-header">
+        <div className="card-header">
           <b>Server information</b>
         </div>
         <div className="card-body">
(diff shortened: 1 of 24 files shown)


src/components/HomePage.tsx
```diff
@@ -47,7 +47,7 @@ function SessionInfo({ session: { serverInfo } }) {
   return (
     <div>
       <div className="card server-info-panel">
-        <div className="card-header">
+        <div className="card-header" data-testid="home-header">
```
grahamalama, line 50: I remember the Testing Library [docs](https://testing-library.com/docs/queries/about/#priority) that they recommend:
> 1. Queries Accessible to Everyone
> 2. Semantic Queries
> 3. Test IDs

Some of the places where we've used test IDs make sense, but I wonder if we could get away with just querying text sometimes 🤔. For example, in the test that uses this test id, could we look for `Server information` text instead?

alexcottner: Good call, and we've done that in many other places along with getting inputs based on label text. But clearly  I was on autopilot when I did this one. Adjusting now.

grahamalama: Nice. That still leaves a few test ids in this PR -- were those cases where a test id is the most pragmatic query to use?

leplatrem (approved):
I find that the changes are sound.

I don't think I feel comfortable blocking this because we abused usage of `data-testid` attributes instead of using text/label/aria values. I would of course be in favor of respecting the testing recommendations, and if we want to do so, then do it now, while we're at it. So I'll delegate final review to @grahamalama 🙏

src/components/HomePage.tsx
alexcottner: Yes, my first attempt was to retrieve an element based on text. Usually that was in a label or the text of a button, but sometimes it was from static text that was a part of that field.  Where that wasn't possible because content was too dynamic, non-existant (like the rjsf form wrapper), or the could be in multiple places, then I switched to a test id's.

This probably isn't perfect. But it should be much better and cause our tests to break less when we upgrade dependencies.

grahamalama: As long as we have good reason to use the escape hatch, I'm happy using the escape hatch 👍🏻

grahamalama (approved):

PR #3104, opened 2024-01-12
Migrate from Webpack and Jest to Vite and Vitest

grahamalama:
This PR all started when I was investigating https://github.com/mozilla/remote-settings/issues/522

I suspected some rerendering was causing elements to go stale, so I fired up Kinto Admin and went to look at the profiler only to discover that React thought we were running a production build.

image

I went to make the necessary configuration changes to enable a development build, but I quickly became disheartened by the task of modifying Webpack config.

I then though about how Webpack is pretty slow in general, compared to newer tools. I tried out a few different bundlers (esbuild, parcel, Vite), and Vite seemed to be the easiest to use and was faster that Webpack.

I made that change, but then quickly found out that `Jest` is incompatible with Vite. I searched for answers there as well and found https://vitest.dev/, which is the "part 2" of this PR. Migrating to Vitest also seemed to expose some errors in our tests

Overall after this PR, we have:
- a faster, more-featured local development experience
- less config to manage

Some of the changes in this PR required mass find-and-replaces (d84f5f3d760eec1aef052706fb92660f29296712 and e0d181597d5f26c530241a4e10b45bfd24fbcd9c), so I'd recommend stepping through by commit.


tsconfig.json
grahamalama, line 1: This was mostly me trying to merge our old tsconfig with what was provided by [their template](https://github.com/vitejs/vite/blob/ed56d96cec8fd3b9b6dae88d5511e386177e6d5c/packages/create-vite/template-react-ts/tsconfig.json)

alexcottner (approved):
It makes me so happy that can swap out testing frameworks this easily now.

grahamalama:
Note that this PR changes one of the environment variables used when building the bundle --

```diff
- SINGLE_SERVER
+ KINTO_ADMIN_SINGLE_SERVER
```

we'll need to account for this in the build process. `ASSET_PATH` should stay the same.

PR #3121, opened 2024-01-16
Resolves #3120 : Exclude groups from loaded permissions

### LINKED ISSUE #3120: Performance issue with a lot of groups
Hello,

We are facing a performance issue with the loading of the Kinto-admin menu when a large number of groups are created (around 5000).
When the home page is loaded, a request loads all the permissions of the current user in order to display the menu.
Records' permissions are excluded from this loading, but not groups' permissions (https://github.com/Kinto/kinto-admin/blob/8ade740afce004553b82b76a652fbe87b091491d/src/sagas/session.ts#L324C13-L324C13).

As a result, the menu is only displayed after a certain amount of time (around 2 seconds per paginated request, with 5000 groups and a pagination of 100 -> around 2 minutes).

Best regards.

slafon51:
This PR resoles #3120 by excluding groups from loaded permissions.

### DIFF
--- a/src/sagas/session.ts
+++ b/src/sagas/session.ts
@@ -321,7 +321,7 @@ export function* listBuckets(
     if ("permissions_endpoint" in serverCapabilities) {
       const { data: permissions } = yield call(
         [client, client.listPermissions],
-        { pages: Infinity, filters: { exclude_resource_name: "record" } }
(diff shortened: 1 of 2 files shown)


alexcottner (approved):
Thank you for the issue and PR, merging this in so it's in the next release.

Issue #3123, opened 2024-01-16
When open-id auth token expires, the system tries to use it anyway upon first load


alexcottner:
Since switching from session-storage to local-storage for auth token, users have reported that they get an "authentication failed" error and sometimes have trouble logging back in properly.

The easy solution would be to discard the auth token if it's expired.

PR #3124, opened 2024-01-16
Fix login screen notifications when token is expired

alexcottner:
Resolves https://github.com/Kinto/kinto-admin/issues/3123 . Adding logic to re-authenticate when users land on the home page with an expired open-id token. Adjusted tests.

This would also be easy to expand and handle other auth types if we want.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
   "scripts": {
     "build": "vite build",
     "build:readme": "toctoc README.md -w",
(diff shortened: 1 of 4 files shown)


leplatrem (approved):
Don't we want a test that verifies that `setupSession` is not called when token is expired?
This maybe the commit title in the end... fix login screen notifications when token is expired?

alexcottner:
I missed that branch in my tests, added that. Also updated the title to your suggestion as that's cleaner.

PR #3125, opened 2024-01-17
Remove unused packages

grahamalama:
This is just a spot check of removing packages that aren't used anymore. Feel free to suggest more if I missed any


alexcottner:
Was gonna make some suggestions but figured it was just as fast to update package.json myself. What do you think?

alexcottner (approved):

PR #3127, opened 2024-01-17
Fixes #3079 - Adding display fields to diff view headers in review process

### LINKED ISSUE #3079: Reviewing small changes to large records is difficult
Reviewing small changes to large records is difficult with the current simple-review page. Example screenshot:
image

We don't know what records are actually being modified without clicking "Show all lines", however if the records are too tall this results in a lot of scrolling. 

If we show the columns from the record list view in the header, users will have a much easier time understanding what they are reviewing without needing the full context of the record.  See example mockup below:
image

alexcottner:
Fixes #3079 - Adding displayFields properties and values to record headers in diff view.
Added unit test to check

image

**Note to reviewers:** I am very open to formatting suggestions for making this more readable.

Screenshot with long properties in the header bar:
image

### DIFF
--- a/src/components/signoff/SimpleReview/PerRecordDiffView.tsx
+++ b/src/components/signoff/SimpleReview/PerRecordDiffView.tsx
@@ -178,12 +178,20 @@ export function formatDiffHeader({
   let fields = [];
 
   for (let f of displayFields) {
-    fields.push(`${f}: ${(target || source)[f] || "undefined"}`);
+    fields.push(
+      <span>
+        <label>{f}:</label> {(target || source)[f] || "undefined"}
+      </span>
+    );
   }
 
-  fields.push((target || source).id);
+  fields.push(
+    <span>
+      <label>id:</label> {(target || source).id}
+    </span>
+  );
 
-  return fields.join(" | ");
+  return <>{fields}</>;
 }
 
 function recordsAreDifferent(a: ValidRecord, b: ValidRecord): boolean {
--- a/src/components/signoff/SimpleReview/PerRecordDiffView.tsx
+++ b/src/components/signoff/SimpleReview/PerRecordDiffView.tsx
@@ -180,7 +180,7 @@ export function formatDiffHeader({
   for (let f of displayFields) {
     fields.push(
       <span>
-        <label>{f}:</label> { renderDisplayField(target || source, f) }
+        <label>{f}:</label> {renderDisplayField(target || source, f)}
       </span>
     );
   }
--- a/src/components/signoff/SimpleReview/PerRecordDiffView.tsx
+++ b/src/components/signoff/SimpleReview/PerRecordDiffView.tsx
@@ -102,7 +102,7 @@ function Diff({
   target?: ValidRecord;
   className?: string;
   allLines?: boolean;
-  displayFields?: string[]
+  displayFields?: string[];
 }) {
   let diff: string[];
 
@@ -124,7 +124,8 @@ function Diff({
       data-testid="record-diff"
     >
       <div className="card-header">
-        <DiffLabel changeType={changeType} /> {formatDiffHeader({ source, target, displayFields })}
+        <DiffLabel changeType={changeType} />{" "}
+        {formatDiffHeader({ source, target, displayFields })}
       </div>
       <div className="card-body p-0">
         <pre className="json-record json-record-simple-review mb-0">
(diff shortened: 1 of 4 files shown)


src/components/signoff/SimpleReview/index.tsx
```diff
@@ -158,6 +158,7 @@ export default function SimpleReview({
         </div>
       );
     }
+    console.log(collection);
```
leplatrem, line 161: leftover ;)

leplatrem (commented):
How does it look with long field values? (overflow)

Another approach could be to force the inclusion of lines that contain the display fields. For example, like in pseudo-code
```
rows = diff(r1, r2);
for field in displayFields:
    if field not in rows:
         rows.prepend(field, (r1 || r2)[field])
```

src/components/signoff/SimpleReview/index.tsx
alexcottner: Fixed

src/components/signoff/SimpleReview/PerRecordDiffView.tsx
```diff
@@ -160,6 +166,34 @@ function DiffLabel({ changeType }: { changeType: ChangeType }) {
...
+  source: ValidRecord;
+  target: ValidRecord;
+  displayFields: string[];
+}) {
+  let fields = [];
+
+  for (let f of displayFields) {
+    fields.push(
+      <span>
+        <label>{f}:</label> {(target || source)[f] || "undefined"}
```
leplatrem, line 183: Since we support dotted field names to target nested objects (eg. `"book.author.name"`), we should use [`renderDisplayField()`](https://github.com/Kinto/kinto-admin/blob/f201c095c81331f295ac988227bbb88446f3f364/src/utils.tsx#L196C17-L196C35) here.

leplatrem:
Since we are diffing serialized JSON, my suggestion to force certain fields to appear in the diff is a challenging to implement (I just read the source of `utils.diffJson()`), so let's go with this approach.

Before approving I'd like to make sure it renders well when the list of displayField is very long, or contains very long values. And also make sure we support the dotted field notation (see `handleNestedDisplayField()`)

src/components/signoff/SimpleReview/PerRecordDiffView.tsx
alexcottner: Oh, hadn't considered nested properties being displayed and I didn't write code to compensate for that. Updating to use `renderDisplayField` which also takes care of the `undefined` issue in a consistent way.

alexcottner:
Switched to use `renderDisplayField` as suggested, this keeps us consistent.

As for long headers, it was just wrapping (screenshot below). I'm gonna say this is functional but not pretty, so I adjusted styling to give it a max-width and truncate after that (2nd screenshot below).

How it would have looked:
image

How it looks now with truncating:
image

src/components/signoff/SimpleReview/PerRecordDiffView.tsx
```diff
@@ -160,6 +166,34 @@ function DiffLabel({ changeType }: { changeType: ChangeType }) {
...
+  source: ValidRecord;
+  target: ValidRecord;
+  displayFields: string[];
+}) {
+  let fields = [];
+
+  for (let f of displayFields) {
+    fields.push(
+      <span>
+        <label>{f}:</label> {renderDisplayField(target || source, f)}
```
leplatrem, line 183: I have a doubt about this. `target` would be `undefined` when record is deleted?  Would it be a tombstone? (eg. `{id: "xxx", deleted: true}`)

test/components/signoff/SimpleReview/PerRecordDiffView_test.jsx
```diff
@@ -135,3 +136,61 @@ describe("findChangeTypes", () => {
...
+        prop3: "prevVal",
+      },
+      target: { 
+        id: "foo",
+        prop1: "val3",
+        prop2: "val4",
+        // prop3 intentionally undefined
+      },
+      displayFields: [ "prop1", "prop2", "prop3"]
+    })).toBe("prop1: val3prop2: val4prop3: <unknown>id: foo");
```
leplatrem, line 184: Wouldn't we want `prop3: prevVal` if undefined in target? (genuine question, not sure)
```js
const merged = Object.keys(target).reduce((acc, k) => {
       if (target[k]) { acc[k] = target[k] }
       return acc;
   }, source
);
...
renderDisplayField(merged, f);
```

leplatrem (approved):
Looks good. Just one doubt about labels when record is deleted or when property is removed from new record

grahamalama:
(Dismissed my review request since it was approved by @leplatrem ✅)

src/components/signoff/SimpleReview/PerRecordDiffView.tsx
alexcottner: If the record was deleted, target would be undefined (or at the very least nullish) so we would render what the source object had for the property.

alexcottner: Example screenshot
image

test/components/signoff/SimpleReview/PerRecordDiffView_test.jsx
alexcottner: I thought about that and I don't think so, we can easily coalesce to the property rather than the object to do that (or do a little spread magic). This way we're keeping our behavior consistent with the list view by always rendering the current property values, even when undefined. 

We can always switch this later if the users prefer something else.

PR #3136, opened 2024-01-22
Combine comment dialog components

grahamalama:
I noticed that we had two remarkably similar `<CommentDialog />` comments, so in this PR, I deleted one and tweaked the other so that it had a compatible API for both places they are used.


alexcottner (approved):
less code = more better

PR #3138, opened 2024-01-23
Enforce import organization with Prettier

grahamalama:
This PR adds a [prettier plugin](https://www.npmjs.com/package/@trivago/prettier-plugin-sort-imports) to enforce import ordering, a la `ruff` or `isort` in our Python projects


.git-blame-ignore-revs
```diff
@@ -0,0 +1,2 @@
+# Run prettier after installing @trivago/prettier-plugin-sort-imports
+f993f8379304ce0f44cae0bd4b76b004aa1caa30
```
leplatrem, line 2: 😎

leplatrem (approved):
💯

PR #3139, opened 2024-01-23
Clean up a few things in tests 

grahamalama:
- remove unused mocks
- use typescript with our test utils module, and make camelCase while we're at it
- Delete duplicate createComponent util
- Tweak renderWithProvider to look like what's suggested in [testing-library](https://testing-library.com/docs/react-testing-library/setup/#custom-render) / [redux](https://redux.js.org/usage/writing-tests#setting-up-a-reusable-test-render-function) docs
- instead of doing
```
const node = render(<Component />)
expect(node.query(...))
```
do
```
render(<Component />)
expect(screen.query(...))
```
as is demonstrated throughout the Testing Library docs


alexcottner (approved):
Did a little reading and it looks like this is the more correct way of doing things. Are we intentionally leaving the re-render logic in there for authForm tests for now? I am curious as to how replacing that would look.

grahamalama:
> Are we intentionally leaving the re-render logic in there for authForm tests for now? I am curious as to how replacing that would look.

Yeah, for now I only did replacements where dropping in `screen` was relatively straightforward. There were a few special cases that probably deserve to be looked at in another PR

Issue #3140, opened 2024-01-25
Enable stricter Eslint checks


grahamalama:
Add:
- `plugin:react/recommended`
- `plugin:react/jsx-runtime`

Enable:
- `@typescript-eslint/recommended-type-checked`
- Any rules that we turned `off`
and maybe
- `@typescript-eslint/stylistic-type-checked`

Issue #2132, opened 2021-10-27
Enable strict mode in Typescript


grahamalama:
As discussed in #2119, enabling [strict](https://www.typescriptlang.org/tsconfig#strict) mode in the Typescript compiler would reduce the chance of certain types of bugs occurring.

Running `tsc --strict` reports 321 errors. Trying to fix these errors in one PR would be a bit daunting. Instead, there might be a way to gradually enable strict checking on a per file/directory basis. There doesn't seem to be a Typescript-official way to do this, but [this package](https://github.com/allegro/typescript-strict-plugin) looks like it could be helpful.

grahamalama:
Another option to bring the codebase up to `strict` compliance would be to enable the compiler rules that comprise strict mode one at at time. Those are:
```
strictNullChecks
strictBindCallApply
strictFunctionTypes
strictPropertyInitialization
noImplicitAny
noImplicitThis
useUnknownInCatchVariables
```

leplatrem:
That's a good idea! Let's start a branch and use this list as checkboxes. I'm willing to help!

PR #3141, opened 2024-01-25
Migrate all remaining JS files to TS, tweak ESLint and Typescript configuration

grahamalama:
This PR:

- Tweaks our ESLint and Typescript configuration to be more line with the [Vite](https://github.com/vitejs/vite/tree/main/packages/create-vite/template-react-ts) and [Vitest](https://github.com/vitest-dev/vitest/tree/main/examples/react-testing-lib) examples
- Migrates all remaining `{js, jsx}` files to `{ts, tsx}`
- Removes some cruft 

In playing around with this configuration, I discovered that we're missing a lot of the type safety that Typescript should provide as well as some important React rules. I turned of some ESLint rules in this PR with the plan of re-enabling them in future PRs. See #3140 and #2132


alexcottner (approved):
# 🛳️

PR #3143, opened 2024-01-26
Set `@src` and `@test` aliases for `src` and `test` directories, respectively

grahamalama:
Now, instead of `../../../<relative imports>`, we can use `@src/` and `@test/`. Just a little quality of life improvement.

8c3de1a7d1ca64461bb58f114842a6448ecf2e4f adds the aliases, and 5415b2d8899a84cdf40915835a8605d2b1b4d2a8 uses the alias, so it's pretty noisy, which is why I added it to the ignore-revs file


alexcottner (approved):

leplatrem (approved):
💯

PR #3148, opened 2024-01-31
Provide `VERSION` file with asset bundle, read in application

grahamalama:
This PR adds a script to generate a `VERSION` file that gets packaged with our built asset bundle. This is so that applications can read that file to report the Kinto Admin version in a Kinto capabilities object

```
> npm run build
> tree build -L 1
build
├── VERSION
├── assets
└── index.html
```


bin/generate-version.sh
```diff
@@ -0,0 +1,12 @@
+#! /bin/bash
+
+CURL_LATEST_RELEASE_VERSION="curl -s https://api.github.com/repos/Kinto/kinto-admin/releases/latest | jq -r .tag_name"
+KINTO_ADMIN_VERSION_FALLBACK=$(git describe --tags --abbrev=4) || eval "$CURL_LATEST_RELEASE_VERSION"
+
+if [ -n "$KINTO_ADMIN_VERSION" ]; then
+    VERSION="${KINTO_ADMIN_VERSION}"
+else
```
leplatrem, line 8: When does this happen? Won't there always be a _latest_ version on the releases page?

grahamalama: > Won't there always be a latest version on the releases page?

Yes, but we want to account for the case where someone wants to build a version that isn't the latest version.

This would be useful in a script like one of our `build-kinto-admin` scripts. Since there's no `git` directory when we `curl -OL "https://github.com/Kinto/kinto-admin/archive/refs/tags/${TAG}.tar.gz"`, having this env var will allow us to inject the correct version (since it won't be able to read from the tag).

bin/generate-version.sh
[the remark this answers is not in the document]
alexcottner: I suppose requiring WSL to build a UI is too much.

package.json
```diff
@@ -4,16 +4,17 @@
   "description": "Kinto Web Administration Console in React.js",
   "type": "module",
   "scripts": {
-    "build": "vite build",
+    "build": "npm run generate-version && vite build",
```
grahamalama, line 7: This seems like a job for a `Makefile`, which can come in another PR. It'd be nice to add our common targets like `build`, `start`, `lint`, `format`, etc. while we're at it.

public/.gitkeep
grahamalama, line 1: If/when we move a "real" file into here, we can ditch this `.gitkeep` file

bin/generate-version.js
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env node
+import { execSync } from "child_process";
+import { writeFileSync } from "fs";
+import * as https from "https";
+
+async function getLatestRelease() {
```
alexcottner, line 6: I think this would be the more "node way" of doing things. But honestly I'm never a fan of just following established patterns, I always prefer what is most readable unless there is a performance issue with that.

```
#!/usr/bin/env node
#!/usr/bin/env node
import { execSync } from "child_process";
import { writeFileSync } from "fs";
import fetch from "node-fetch"; // i think fetch is the most common package to use, axios is good too

async function getLatestRelease() {
  const res = await fetch("https://api.github.com/repos/Kinto/kinto-admin/releases/latest");
  const body = await res.json();
  return body.tag_name;
}

function getGitLatestReleaseVersion() {
  try {
    return execSync("git describe --tags --abbrev=4", {
      encoding: "utf-8",
    }).trim();
  } catch (error) {
    console.log(error);
  }
};

(async() => {
  const kintoAdminVersion = process.env.KINTO_ADMIN_VERSION // coalesce until we get something
    || getGitLatestReleaseVersion() 
    || await getLatestRelease();
  writeFileSync("./public/VERSION", kintoAdminVersion);
})();
```

Note: This requires the `node-fetch` package as a dev dependency.

package.json
alexcottner: I don't think I've ever seen a Makefile for a pure react project before. I'm not opposed to this, but it might cause friction with pure UI devs. We should ask around a bit.

bin/generate-version.js
grahamalama: > Note: This requires the node-fetch package as a dev dependency.

At first I was trying to avoid downloading `node-fetch`, but since people need to `node ci` before running `node run build` anyway, what's one more dependency?

alexcottner (approved):

bin/generate-version.js
alexcottner: I try to keep dependencies slim, but when it comes to things like http clients or database clients I think they're worth it.

PR #3178, opened 2024-03-01
Resolves issue #3159 - records list view does not crash on anonymous users

### LINKED ISSUE #3159: Crash in `RecordTable.tsx:21` when navigating to read-only collection
**Steps to reproduce**

1. `npm start` on d26f2c36ed04872dbdb6aae5a9c46efc2c916262
1. Use server `https://remote-settings-dev.allizom.org/v1/`
2. Authentication method: `Anonymous`
3. Navigate to `#/buckets/main-workspace/collections/normandy-recipes/records`

**Expected**

Would open records of collection in read-only mode

**Actual**

Crashes with:

```
Uncaught TypeError: bucket.data is undefined
    ListActions RecordTable.tsx:21
    React 12
    unstable_runWithPriority scheduler.development.js:468
    React 4
    Redux 32
    updateRoute routes.tsx:18
    ComponentWrapper routes.tsx:22
    React 5
    unstable_runWithPriority scheduler.development.js:468
    React 4
    unstable_runWithPriority scheduler.development.js:468
    React 17
    <anonymous> index.tsx:5
2 RecordTable.tsx:21:32

``

### PR DESCRIPTION
**Pull request by alexcottner**
Resolves #3159 . Added unit test to prevent us from making the same mistake again.

This also relates to #3075. I think the solution fixes the UI rendering issues, but things can still end up in a bad state in redux. But I can't find any symptoms of the problem anymore.

### CODE REVIEW

File: src/components/signoff/SignoffToolBar.tsx [hunk 1]
```diff
@@ -79,6 +79,10 @@ export default function SignoffToolBar({
     }
   }, [collectionState.data]);
 
+  if (!bucketState.data?.id) {
```
leplatrem, line 82: When does this happen exactly? Or in other words, why navigating to a read-only bucket leads to bucket data being empty?

If I navigate to `/buckets/bid/collections/cid`, even if I don't have permissions to see the bucket details, the bucket data could at least be `{"data": {"id": "bid"}}`

test/components/CollectionRecords_test.tsx
```diff
@@ -272,4 +272,42 @@ describe("CollectionRecords component", () => {
...
+        data: null,
+      },
+      collection: {
+        busy: false,
+        data: {
+          schema: {
+            type: "object",
+            properties: {
+              foo: { type: "string" },
+            },
```
leplatrem, line 292: Is that a realistic situation? i.e. having collection data but no bucket data?

src/components/signoff/SignoffToolBar.tsx
alexcottner: Replied in other comment, but this only happens when a user can view record data but not the collection attributes.

leplatrem: Ok, I see. So this patch hides the signoff toolbar if no permissions. Makes sense.

leplatrem (approved):

PR #3180, opened 2024-03-01
Major upgrades of dependencies

alexcottner:
Decided to do some upgrades of several packages, some are major version bumps (like react).
Made bugfixes/updates for problems I could identify.
Everything appears to be working well, but maybe we should do some more manual testing just to be sure.


test/components/AuthForm_test.tsx
```diff
@@ -35,7 +35,8 @@ describe("AuthForm component", () => {
       setupSession = vi.fn();
       serverChange = vi.fn();
       (getServerInfo = async () => {
-        new Promise(resolve => setTimeout(resolve, 1000)); // simulate server response taking a second
+        console.log("doing stuff");
```
leplatrem, line 38: leftover ?

test/components/Layout_test.tsx
```diff
@@ -17,20 +17,21 @@ describe("App component", () => {
...
 
-    it("should render a session top bar when anonymous", () => {
+    it("should render a session top bar when anonymous", async () => {
       const serverInfo = {
         url: "http://test.server/v1/",
         capabilities: {},
       };
       store.dispatch(sessionActions.serverInfoSuccess(serverInfo));
       store.dispatch(sessionActions.setAuthenticated());
+      await waitFor(() => new Promise(resolve => setTimeout(resolve, 500))); // debounce wait
```
leplatrem, line 27: What does this do here concretely?

test/components/Layout_test.tsx
```diff
@@ -59,6 +61,7 @@ describe("App component", () => {
       store.dispatch(sessionActions.serverInfoSuccess(serverInfo));
       store.dispatch(sessionActions.setupComplete());
       store.dispatch(sessionActions.setAuthenticated(credentials));
+      await waitFor(() => new Promise(resolve => setTimeout(resolve, 500))); // debounce wait
```
leplatrem, line 64: Does `setAuthenticated()` take a few hundreds of msec so that this is necessary?

leplatrem (approved):

test/components/Layout_test.tsx
alexcottner: Changing authentication contexts takes a little bit of time between the ServerHistory [debounced fetch call](https://github.com/Kinto/kinto-admin/blob/master/src/components/ServerHistory.tsx#L72) which has 400ms of waiting, and redux.

It seems like with the upgrade to react 18 our test weren't waiting automatically in some cases. This should only be a problem on tests that hit `AuthForm` and `HomePage` (which includes AuthForm) without mocking session up front.

alexcottner:
Had a nasty merge conflict in package-lock.json, re-applied changes and force pushed to make commit history cleaner.

PR #3207, opened 2024-04-05
Fixes #3202: allowing self review when `to_review_enabled` is false

### LINKED ISSUE #3202: Always show `Approve`/`Reject` buttons when server has review disabled
When review is disabled, users can approve their own changes. But the Admin UI does not show the approve/reject buttons.

The only way for them to do it is to use the CLI:

```
$ echo '{"data":{"status": "to-resign"}}' | http PATCH https://remote-settings-dev.allizom.org/v1/buckets/main-workspace/collections/product-integrity "Authorization:$BEARER_TOKEN"
```

alexcottner:
Resolves #3202 - allowing for self review when `to_review_enabled` is false on either the signer plugin level or the specific resource mapping. 

Added unit tests to simple-review and legacy review toolbar to ensure consistent behavior.

### DIFF
--- a/src/components/signoff/SignoffToolBar.tsx
+++ b/src/components/signoff/SignoffToolBar.tsx
@@ -110,10 +110,12 @@ export default function SignoffToolBar({
   const { status } = source;
 
   const canRequestReview = canEdit && isEditor(source, sessionState);
+
   const canReview =
     canEdit &&
-    isReviewer(source, sessionState) &&
-    !hasRequestedReview(source, sessionState);
+    ((isReviewer(source, sessionState) &&
+      !hasRequestedReview(source, sessionState)) ||
+      !sessionState.serverInfo?.capabilities?.signer?.to_review_enabled);
   const canRollback = canEdit;
   const hasHistory = "history" in sessionState.serverInfo.capabilities;
 
--- a/src/components/signoff/SignoffToolBar.tsx
+++ b/src/components/signoff/SignoffToolBar.tsx
@@ -111,11 +111,26 @@ export default function SignoffToolBar({
 
   const canRequestReview = canEdit && isEditor(source, sessionState);
 
+  let toReviewEnabled =
+    sessionState.serverInfo?.capabilities?.signer?.to_review_enabled === true;
+  if (toReviewEnabled) {
+    const resourceMatch =
+      sessionState.serverInfo?.capabilities?.signer?.resources?.find(
+        x =>
+          x.source.bucket === source.bid &&
+          x.source.collection === source.cid &&
+          x.destination.bucket === destination.bid &&
+          x.destination.collection === destination.cid
+      );
+    toReviewEnabled =
+      !resourceMatch || resourceMatch.to_review_enabled !== false;
+  }
+
   const canReview =
     canEdit &&
     ((isReviewer(source, sessionState) &&
       !hasRequestedReview(source, sessionState)) ||
-      !sessionState.serverInfo?.capabilities?.signer?.to_review_enabled);
+      !toReviewEnabled);
   const canRollback = canEdit;
   const hasHistory = "history" in sessionState.serverInfo.capabilities;
 
--- a/src/components/signoff/SignoffToolBar.tsx
+++ b/src/components/signoff/SignoffToolBar.tsx
@@ -5,7 +5,7 @@ import HumanDate from "./HumanDate";
 import { ProgressBar, ProgressStep } from "./ProgressBar";
 import { DiffInfo, Review } from "./Review";
 import { Signed } from "./Signed";
-import { isMember } from "./utils";
+import { isMember, toReviewEnabled } from "./utils";
 import { canEditCollection } from "@src/permission";
 import type {
   BucketState,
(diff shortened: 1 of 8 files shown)


src/components/signoff/SignoffToolBar.tsx
```diff
@@ -110,10 +110,12 @@ export default function SignoffToolBar({
...
 
   const canRequestReview = canEdit && isEditor(source, sessionState);
+
   const canReview =
     canEdit &&
-    isReviewer(source, sessionState) &&
-    !hasRequestedReview(source, sessionState);
+    ((isReviewer(source, sessionState) &&
+      !hasRequestedReview(source, sessionState)) ||
+      !sessionState.serverInfo?.capabilities?.signer?.to_review_enabled);
```
leplatrem, line 118: If we want to be thorough, we may want to align this test with the possibility of the server. 

Indeed, the global `to_review_enabled` can be true at the server level (which is the case of the DEV server for example), and be set to false for a specific collection. 
For example, in our prod config we have:
```ini
    ### Normandy recipes
    # Only the Normandy server is allowed to publish recipes. Multi-signoff is disabled.
    kinto.signer.main-workspace.normandy-recipes.to_review_enabled = false
    kinto.signer.main-workspace.normandy-recipes.group_check_enabled = false
    kinto.signer.main-workspace.normandy-recipes-capabilities.to_review_enabled = false
    kinto.signer.main-workspace.normandy-recipes-capabilities.group_check_enabled = false
```
which then ends up in `serverInfo.capabilities.signer.resources[{i}].to_review_enabled` as you can see here https://remote-settings.allizom.org/v1/ 
```
"signer": {
      "description": "Digital signatures for integrity and authenticity of records.",
      "url": "https://github.com/Kinto/kinto-signer#kinto-signer",
      "version": "32.1.0",
      "resources": [
        {
          "source": {
            "bucket": "staging",
            "collection": null
          },
          "destination": {
            "bucket": "blocklists",
            "collection": null
          },
          "preview": {
            "bucket": "blocklists-preview",
            "collection": null
          }
        },
        {
          "source": {
            "bucket": "main-workspace",
            "collection": "normandy-recipes"
          },
          "destination": {
            "bucket": "main",
            "collection": "normandy-recipes"
          },
          "to_review_enabled": false
        },
```

Also I wouldn't test for falsy, I would test that the `to_review_enabled` is explicitly set to false

leplatrem (commented):
Thanks for starting this! 

We may have to write a util that explicitly looks for the `to_review_enabled` flag in the collection, bucket, and server wide...

src/components/signoff/SignoffToolBar.tsx
alexcottner: Ooo, didn't realize there was a possibility of multiple layers here.

A big follow up question (since I don't see existing unit tests verifying this functionality). If I have defined values for `to_review_enabled` at both the server level and collection level, which one "wins"? Does the collection level always win? Or if either are `false`, do we always consider it disabled?

leplatrem: So, if we want the client behaviour to match the server's, then _collection overrides the server_.

I don't think it would affect your code, but note that in practice, we use it only in one way: the server has review enabled, and we disable it for certain collections. We don't use the opposite, like for example on the dev server review is globally disabled but we don't enable it for certain collections.

alexcottner: Got it. So putting this into pseudo code logic:
```
review_is_enabled = server_to_review_enabled && collection_to_review_enabled
```

leplatrem: > `review_is_enabled = server_to_review_enabled && collection_to_review_enabled`
>

Yes, except that `collection_to_review_enabled` is implicitly true

alexcottner: I _think_ I've captured this logic in the latest changes.

src/components/signoff/SimpleReview/index.tsx
```diff
@@ -65,9 +65,25 @@ export default function SimpleReview({
...
+        x =>
+          x.source.bucket === sourceBid &&
+          x.source.collection === sourceCid &&
+          x.destination.bucket === destBid &&
+          x.destination.collection === destCid
+      );
+    toReviewEnabled =
+      !resourceMatch || resourceMatch.to_review_enabled !== false;
+  }
+
```
leplatrem, line 82: This could be moved into https://github.com/Kinto/kinto-admin/blob/master/src/components/signoff/utils.ts instead of duplicating

src/components/signoff/SignoffToolBar.tsx
```diff
@@ -110,10 +110,27 @@ export default function SignoffToolBar({
...
+      sessionState.serverInfo?.capabilities?.signer?.resources?.find(
+        x =>
+          x.source.bucket === source.bid &&
+          x.source.collection === source.cid &&
+          x.destination.bucket === destination.bid &&
+          x.destination.collection === destination.cid
+      );
+    toReviewEnabled =
+      !resourceMatch || resourceMatch.to_review_enabled !== false;
+  }
```
leplatrem, line 127: There is one case missing: when review is enabled/disabled at the bucket level:
```js
if (!resourceMatch) {
  // No specific config for this collection.
  // Check at bucket level.
  resourceMatch =
    sessionState.serverInfo?.capabilities?.signer?.resources?.find(
      x =>
        x.source.bucket === source.bid &&
        x.source.collection === null &&
        x.destination.bucket === destination.bid &&
        x.destination.collection === null
    );
}
```

But it is not used in practice, so I guess we can leave it like this. Maybe add a code comment.

Example in python:
https://github.com/mozilla-services/remote-settings-lambdas/blob/5c261be0bd927a0ce8af79f045c9eab4ae23a9d2/commands/backport_records.py#L102-L126

src/components/signoff/SignoffToolBar.tsx
```diff
@@ -110,10 +110,27 @@ export default function SignoffToolBar({
...
+    const resourceMatch =
+      sessionState.serverInfo?.capabilities?.signer?.resources?.find(
+        x =>
+          x.source.bucket === source.bid &&
+          x.source.collection === source.cid &&
+          x.destination.bucket === destination.bid &&
+          x.destination.collection === destination.cid
+      );
+    toReviewEnabled =
+      !resourceMatch || resourceMatch.to_review_enabled !== false;
```
leplatrem, line 126: I think it would be clearer to write:
```js
if (resourceMatch && resourceMatch.hasOwnProperty(to_review_enabled)) {
    toReviewEnabled = resourceMatch.to_review_enabled;
}
```

leplatrem (approved):
I think what you have now is correct. But I found the code hard to read/verify. Not blocking landing, but please consider taking it into account if my feedback makes sense

src/components/signoff/SignoffToolBar.tsx
alexcottner: Ended up refactoring this a bit to be easier to read.

alexcottner:
> I think what you have now is correct. But I found the code hard to read/verify. Not blocking landing, but please consider taking it into account if my feedback makes sense

I did a a little refactor to make it more reasonable to test and read.

PR #3218, opened 2024-04-15
Switching auth form spinner to be based on session.busy

alexcottner:
Small fix for issue that we noticed after v3.0.7 was created. It looks like rjsf may have switched the order of events of form onChange vs field onChange. After looking at the code a bit I realized we should be indifferent to this.

Adjusted code to show a spinner when `session.busy` is true. Adjusted unit tests accordingly.

### DIFF
--- a/test/components/AuthForm_test.tsx
+++ b/test/components/AuthForm_test.tsx
@@ -77,9 +77,7 @@ describe("AuthForm component", () => {
         fireEvent.change(screen.queryByLabelText("Server*"), {
           target: { value: "http://test.server/v1" },
         });
-        await screen.findByTestId("spinner"); // spinner should show up
         await waitFor(() => new Promise(resolve => setTimeout(resolve, 500))); // debounce wait
-        expect(screen.queryByTestId("spinner")).toBeNull(); // spinner should be gone by now
 
         fireEvent.click(screen.getByLabelText("Basic Auth"));
         fireEvent.change(screen.getByLabelText("Username*"), {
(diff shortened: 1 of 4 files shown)


test/components/AuthForm_test.tsx
```diff
@@ -77,9 +78,7 @@ describe("AuthForm component", () => {
         fireEvent.change(screen.queryByLabelText("Server*"), {
           target: { value: "http://test.server/v1" },
         });
-        await screen.findByTestId("spinner"); // spinner should show up
         await waitFor(() => new Promise(resolve => setTimeout(resolve, 500))); // debounce wait
-        expect(screen.queryByTestId("spinner")).toBeNull(); // spinner should be gone by now
```
grahamalama, line 82: Don't we still want to make some assertions about a spinner appearing then disappearing?

src/reducers/session.ts
```diff
@@ -68,9 +69,12 @@ export default function session(
         }),
       };
     }
+    case SESSION_GET_SERVERINFO: {
```
grahamalama, line 72: How is this similar / different to `SESSION_BUCKETS_REQUEST` and `SESSION_BUSY`?

Maybe what was missing here was that we forgot to call `sessionBusy(<bool>)` in the `getServerInfo` saga?

grahamalama (commented):
Out of scope for this PR, but could this mean there are other places where our `const [showSpinner, setshowSpinner] = useState(false);` pattern could be competing with some Redux `busy` prop?

src/reducers/session.ts
alexcottner: Hmm, that is a bit cleaner. Switching to that.

grahamalama (approved):

Issue #253, opened 2016-10-04
High level validation of JSON schema


leplatrem:
It would be nice to have some sanity check for the JSON schema + ui:schema, for example the required fields consistency... Otherwise we have to go to the create record form to see the errors. 
Maybe a *preview panel* would help ?

n1k0:
i'm planning to work on that after #371 lands. Basically, JSONEditor should perform JSON validation, to their consumers/wrappers.

anselmos:
@leplatrem  - The #371 has been merged, is there any possibility for you to work on it? Or maybe there is anything that I could do to help with this ?

leplatrem:
Unfortunately, I can't guarantee that this issue will be prioritized anytime soon on our side.

But if you're willing to work on this, it would be excellent! And I would be glad to guide you or review your contribution!

Thanks for your interest! Let us know :)

PR #3232, opened 2024-05-03
Fix #625: Show an error message when rjsf component crashes

### LINKED ISSUE #625: Page crash if user sets wrong ui:widget
screenshot from 2018-09-27 18-18-02

It should probably be better to show an error within the UI

alexcottner:
Resolves #625 - Show an error message when rjsf component crashes due to bad config.
Added comments detailing why we're doing things this way and unit test to verify it works as expected.

image

### DIFF
--- a/src/components/BaseForm.tsx
+++ b/src/components/BaseForm.tsx
@@ -12,6 +12,7 @@ const FormWithTheme = withTheme(Bootstrap4Theme);
 
 export type BaseFormProps = Omit<FormProps, "validator"> & {
   showSpinner?: boolean;
+  formCrashMsg?: ReactNode;
   onSubmit: (data: RJSFSchema) => void;
 };
 
@@ -37,7 +38,7 @@ export default function BaseForm(props: BaseFormProps) {
 
   return (
     <div className="formWrapper" ref={formRef} data-testid="formWrapper">
-      <ErrorBoundary>
+      <ErrorBoundary formCrashMsg={props.formCrashMsg}>
         <FormWithTheme
           {...restProps}
           focusOnFirstError={errorFocus}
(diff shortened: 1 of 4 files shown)


src/components/BaseForm.tsx
```diff
@@ -37,17 +37,65 @@ export default function BaseForm(props: BaseFormProps) {
...
+  declare props: {
+    children: ReactNode;
+  };
+
+  constructor(props) {
+    super(props);
+    this.state = {};
+  }
+
+  static getDerivedStateFromError(thrown) {
```
leplatrem, line 77: I see that this is the key method in this pattern https://react.dev/reference/react/Component#static-getderivedstatefromerror

src/components/BaseForm.tsx
```diff
@@ -37,17 +37,65 @@ export default function BaseForm(props: BaseFormProps) {
...
+  }
+
+  render() {
+    if (this.state.thrown) {
+      return (
+        <>
+          <h2>Error rendering form</h2>
+          <div>
+            This is likely caused by a bad <code>ui:widget</code> value in this
+            collection's UI schema.
```
leplatrem, line 90: Should we make this a link so that user can open collection attributes and fix it?

leplatrem (approved):
👌 

Wrapping the component seems a reasonable pattern.
Ideally #253 could help, but errors in schemas can be introduced via other means than the UI so preventing this is a plus 💯

grahamalama (approved):

src/components/BaseForm.tsx
alexcottner: Added a parameter for custom error messages, adjusted unit test, and adjusted RecordForm and RecordBulk components to send a link with their custom messages. Good call.

PR #3240, opened 2024-05-09
Resolves #3063 - always show the current bucket and collection in the sidebar

### LINKED ISSUE #3063: Always show current bucket/collection in sidebar
Steps to reproduce

* Do not tick `Show readonly buckets/collections`
* Navigate to a read-only collection manually (or from a link, like the preview in the review UI)

Expected

* The current bucket / collection would show up in the sidebar

Actual

* The sidebar does not highlight anything

Screenshot 2023-11-16 at 17 37 55

alexcottner:
Resolves #3063 - always show the current bucket and collection in the sidebar. Added unit tests.
Also removed a rogue console.log.

### DIFF
--- a/test/components/Sidebar_test.tsx
+++ b/test/components/Sidebar_test.tsx
@@ -91,6 +91,48 @@ describe("Sidebar component", () => {
     });
   });
 
+  describe("Read only collections", () => {
+    let bucketMenus;
+
+    beforeEach(() => {
+      renderWithProvider(
+        <Sidebar
+          match={{ params }}
+          location={location}
+          capabilities={capabilities}
+        />,
+        {
+          initialState: {
+            session: {
+              ...session,
+              buckets: session.buckets.map(b => {
+                return {
+                  ...b,
+                  collections: b.collections.map(c => {
+                    return {
+                      ...c,
+                      readonly: true,
+                    };
+                  }),
+                };
+              }),
+            },
+          },
+        }
+      );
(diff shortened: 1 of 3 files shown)


leplatrem (approved):
👍 

I was trying to stretch this a little bit and imagine a situation where we reach a bucket and not a collection (ie. `bid=main` `cid=null` like for `https://remote-settings-dev.allizom.org/v1/admin/#/buckets/main/attributes`) but currently we don't really have anything consistent to show in the sidebar if the bucket's collections list is empty. Plus that's not a common usecase.

:shipit:

test/components/Sidebar_test.tsx
```diff
@@ -91,6 +91,48 @@ describe("Sidebar component", () => {
...
+      bucketMenus = screen.getAllByTestId("sidebar-bucketMenu");
+    });
+
+    it("should highlight the selected collection even if it's read-only and we haven't enabled show read-only collections", () => {
+      const targetEntry = bucketMenus[0].querySelectorAll(
+        ".collections-menu-entry"
+      )[0];
+
+      expect(targetEntry.classList.contains("active")).toBe(true);
+      expect(screen.getAllByTestId("sidebar-menuEntry")).toHaveLength(1); // should only have 1 menu entry for the current read-only collection
```
grahamalama, line 132: Should we use `getByTestId`, since `getBy*` [will](https://testing-library.com/docs/queries/about#types-of-queries)

>  throw a descriptive error if no elements match or if more than one match is found

PR #3252, opened 2024-05-23
Resolves #3220 - show heartbeat status in top bar

### LINKED ISSUE #3220: Show heartbeat status in top bar
Like with a red/green circle and a blank target link to the endpoint

Would serve as a *is it down for everyone or just me* indicator

alexcottner:
Resolves #3220 . Show the heartbeat endpoint status in top bar.  Status icon is also a link that will take the user to the `__heartbeat__` endpoint. Also resolves a brief error that users see during redirect to auth.
- [x] Add check to top bar (first draft done)
- [x] Add unit tests
- [x] Fill in missing unit tests (don't have tests for SessionInfoBar yet)

Screenshots of error (red !) vs success (green circle):
image
Screenshot from 2024-05-28 10-18-53


src/components/SessionInfoBar.tsx
```diff
@@ -34,6 +63,7 @@ export function SessionInfoBar() {
         >
           <Clipboard className="icon" />
         </a>
+        { isHealthy ? <CircleFill color="green" title="Server heartbeat status is healthy" /> : <CircleFill color="red" title="Server heartbeat status IS NOT healthy" /> }
```
alexcottner, line 66: I am slightly color blind and this is messing with me. Might need to do a little gradient to make it easier to see.

alexcottner: Changed icons to make it easier for myself.

src/actions/session.ts
```diff
@@ -17,7 +17,15 @@ import {
 } from "@src/constants";
 import type { ActionType, AuthData, ServerInfo } from "@src/types";
 
-type NavigationResult = ActionType<typeof notifyError> | { type: null };
+const AUTH_REDIRECT_RESULT: ActionType<typeof notifyError> = {
+  notification: {
```
alexcottner, line 21: The brief error users could see when redirecting to the auth provider is due to the methods below returning incomplete action data for notification actions. Providing a simple "Redirection to auth provider..." message seems like the simplest solution.

src/components/SessionInfoBar.tsx
```diff
@@ -10,6 +16,33 @@ export function SessionInfoBar() {
     store => store.session.serverInfo
   );
   const dispatch = useAppDispatch();
+  const [isHealthy, setIsHealthy] = useState(true);
+  const client = getClient();
+
+  const checkHeartbeat = async () => {
+    try {
+      let res: KintoResponse = await client.execute({
```
grahamalama, line 24: Will this actually be a `KintoResponse`? Isn't this more a `Record<string, bool>`? (That probably isn't the right syntax, but you get what I mean)

src/components/SessionInfoBar.tsx
```diff
@@ -10,6 +16,33 @@ export function SessionInfoBar() {
...
+        path: "/__heartbeat__",
+        headers: undefined,
+      });
+      for (let p in res) {
+        if (!res[p]) {
```
grahamalama, line 29: Can we use more descriptive variable names?

src/components/SessionInfoBar.tsx
```diff
@@ -10,6 +16,33 @@ export function SessionInfoBar() {
     store => store.session.serverInfo
   );
   const dispatch = useAppDispatch();
+  const [isHealthy, setIsHealthy] = useState(true);
+  const client = getClient();
+
+  const checkHeartbeat = async () => {
```
grahamalama, line 22: I can see the argument for why this deserves to be local state, but for the sake of consistency, I feel like this should be done through sagas. We only use the client directly like this in 2 components, compared to the many more instances where we use it in sagas.

src/components/SessionInfoBar.tsx
```diff
@@ -1,7 +1,13 @@
...
-import { BoxArrowRight } from "react-bootstrap-icons";
+import { KintoResponse } from "kinto/lib/types";
+import React, { useEffect, useState } from "react";
+import {
+  BoxArrowRight,
+  CircleFill,
+  ExclamationCircleFill,
+} from "react-bootstrap-icons";
 import { QuestionCircleFill } from "react-bootstrap-icons";
 import { Clipboard } from "react-bootstrap-icons";
```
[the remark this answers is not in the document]
alexcottner: Oh silly me, I thought the linter would fix that.

src/components/SessionInfoBar.tsx
alexcottner: Yeah, technically execute will return `any`. But that endpoint in particular should return a hash table that is mostly bools (the details array being the exception). I'll make this more explicit.

src/components/SessionInfoBar.tsx
```diff
@@ -1,7 +1,13 @@
 import * as SessionActions from "@src/actions/session";
+import { getClient } from "@src/client";
 import { useAppDispatch, useAppSelector } from "@src/hooks/app";
-import * as React from "react";
-import { BoxArrowRight } from "react-bootstrap-icons";
+import { KintoResponse } from "kinto/lib/types";
+import React, { useEffect, useState } from "react";
```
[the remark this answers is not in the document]
alexcottner: It would be cool if we could make the linter standardize these. 🤔

src/components/SessionInfoBar.tsx
alexcottner: Ah, I've been slowly pushing towards hooks and away from sagas/redux. But I didn't make this a shared hook since it's only used in one component. We should have a bigger conversation around that. I'm a big proponent of removing redux to reduce dependencies and reduce dependency injection.

grahamalama: > I've been slowly pushing towards hooks and away from sagas/redux

It's worth having a conversation about the way we (ab)use redux + sagas, but until we make a larger decision about how we want to change the way we do state + data fetching, I think it'd be good to keep things internally consistent

alexcottner: Made a first attempt

src/components/SessionInfoBar.tsx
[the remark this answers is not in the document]
grahamalama: Yeah, hopefully there's a ts/eslint option for this

grahamalama (commented):
Thanks for updating this to use Redux + Sagas. I spun up [this issue](https://github.com/Kinto/kinto-admin/issues/3262) for us to track the conversation for adopting new state management / data fetching patterns.

For this PR:
- I see you elected to add a new `heartbeat` slice of the store. Any reason you didn't opt too add a property in the `session` slice, e.g. `session.heartbeat`? We are adding this to the `SessionInfoBar`, after all
- Adding a few saga tests (see [these examples](https://github.com/Kinto/kinto-admin/blob/master/test/sagas/session_test.ts) for session sagas) might be a good idea

alexcottner:
> I see you elected to add a new `heartbeat` slice of the store. Any reason you didn't opt too add a property in the `session` slice, e.g. `session.heartbeat`?

Mainly because the session slice is written to `localStorage`, and that felt a little awkward. Do you think that's a reasonable place for it though?

alexcottner:
Added a few tests for the saga.

src/components/SessionInfoBar.tsx
```diff
@@ -1,15 +1,35 @@
...
   const dispatch = useAppDispatch();
+
+  const checkHeartbeat = async () => {
+    dispatch(HeartbeatActions.heartbeatRequest());
+    setTimeout(checkHeartbeat, 60000);
+  };
+
+  useEffect(() => {
+    checkHeartbeat();
+  }, []);
```
grahamalama, line 31: Does this empty dependency array mean this effect will only fire once? I see the timeout above, but I'm surprised that's not somehow included in the effect to trigger a re-render.

grahamalama (commented):
> Do you think that's a reasonable place for it though?

It does feel more reasonable to have this in the `session` slice, but yeah, doesn't make sense to save to local storage. Maybe we could update `saveSession` / `loadSession` to account for that.

However, seeing as this is blocking a Remote Settings release. Maybe we can call this PR good for now and create a follow on issue.

I think all we need to call this PR done is 1 test that demonstrates that the heartbeat is re-fetched after the timeout winds down.

src/components/SessionInfoBar.tsx
alexcottner: Correct, an empty dependency array will only fire once. If you exclude the parameter, it will run every time. And if you include objects, it will fire whenever they change.

The setTimeout does work properly and re-fire the saga and trigger a re-render every minute as expected. That is due to `useEffect` being a hook.

Issue #3262, opened 2024-06-04
Draft ADR for state management / data fetching strategies throughout app


grahamalama:
We currently use Redux for global state management and Sagas to do data fetching for components. There are new strategies for managing these concerns in React. We should either write an ADR or otherwise document our intentions to migrate to new patterns our maintain our current patterns.

Issue #2133, opened 2021-10-27
Consider Redux Saga Test Plan for saga testing


grahamalama:
To tests sagas, we currently manually advance the saga generator through each step and make assertions about what action is called at each step.

We may want to consider adopting [Redux Saga Test Plan](https://redux-saga-test-plan.jeremyfairbank.com/) for saga testing. Not only does it provide nice sugar for testing sagas step by step as we do now, but it also has nice affordances for integration tests if we want to take that approach.

alexcottner:
closing as we're moving away from redux

PR #3279, opened 2024-06-19
State management change proposal

alexcottner:
Created a proposal to change our state management from redux to hooks.

[Link to ADR for easy reference/rendering](https://github.com/Kinto/kinto-admin/blob/adr-state/docs/adrs/001-state.md)

There are two supporting pull requests showing why I'm proposing this linked in the doc. Here they are again for easy reference:

- #3271 
- #3270 

Affected issues:

- #3262 
- #2135 
- #2133 
- #3416 

Looking for any thoughts/comments/concerns from any active kinto devs.

### DIFF
--- a/docs/adrs/001-state.md
+++ b/docs/adrs/001-state.md
@@ -1,7 +1,7 @@
 # The state of state management in kinto-admin
 
- - Status: proposed
- - Date: 2024-06-19
+ - Status: implementing
+ - Date: 2025-03-20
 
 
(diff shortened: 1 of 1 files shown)


leplatrem:
We probably want this as a blocker for #3416

alexcottner:
Do we want to accept this proposal and start working towards the goal? I can do a PoC PR for one of our components.

leplatrem (approved):

PR #3366, opened 2024-10-14
Fixes #7 - Adds a simple quick search to records list.

### LINKED ISSUE #7: Add live search/filtering to the list of records view.
For quickly finding a listed entry.

alexcottner:
Resolves #7 - adds a simple client-side quick search to the record list view.

There are (obviously) more flexible or advanced options we could use here. And we could implement this server side to better handle very large collections. But if our idea is to do something simple to let users quickly filter down the records for our typical collections, this is a low-code way to meet our needs.

[demo video](https://github.com/user-attachments/assets/9f702bd0-83dc-4bb3-bee3-d93507ffed13)

### DIFF
--- a/src/components/collection/RecordTable.tsx
+++ b/src/components/collection/RecordTable.tsx
@@ -216,13 +216,10 @@ export default function RecordTable({
     <>
       <input
         type="text"
-        className="form-control"
+        className="form-control quickFilter"
         placeholder="Quicksearch"
         onChange={onFilterChange}
         value={filter}
-        style={{
-          marginBottom: "-0.5em",
-        }}
         data-testid="quickFilter"
       />
       <PaginatedTable
--- a/src/components/collection/RecordTable.tsx
+++ b/src/components/collection/RecordTable.tsx
@@ -7,7 +7,7 @@ import SignoffContainer from "@src/containers/signoff/SignoffToolBar";
 import { canCreateRecord } from "@src/permission";
 import type { RecordData } from "@src/types";
 import { capitalize } from "@src/utils";
-import React from "react";
+import React, { useState } from "react";
 import { SortUp } from "react-bootstrap-icons";
 import { SortDown } from "react-bootstrap-icons";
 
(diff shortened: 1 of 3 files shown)


src/components/collection/RecordTable.tsx
```diff
@@ -201,13 +213,26 @@ export default function RecordTable({
...
 
   return (
-    <PaginatedTable
-      thead={thead}
-      tbody={tbody}
-      dataLoaded={recordsLoaded}
-      colSpan={displayFields.length + 2}
-      hasNextPage={hasNextRecords}
-      listNextPage={listNextRecords}
-    />
```
leplatrem, line 211: Couldn't the filter be part of the `PaginatedTable` instead of just records? 
We have list of objects in Remote Settings (eg. `main-workspace` groups or collections) that could definitely benefit from such a quick filter

src/components/collection/RecordTable.tsx
```diff
@@ -201,13 +213,26 @@ export default function RecordTable({
...
+    <>
+      <input
+        type="text"
+        className="form-control"
+        placeholder="Quicksearch"
+        onChange={onFilterChange}
+        value={filter}
+        style={{
+          marginBottom: "-0.5em",
+        }}
```
leplatrem, line 225: Don't we want to push this in CSS with a specific class instead?

leplatrem (approved):
Niiiiiiiice! I like the minimalist code with maximum UX impact 😊 

Just wondering whether we could generalize it instead of just offering the feature in records lists

src/components/collection/RecordTable.tsx
alexcottner: I had thought about that, and we can do that if we refactor our `PaginatedTable` component to do more work. Right now it doesn't handle rendering the records, it just receives a `tbody` property with the rows already pre-rendered. So any filtering it could currentlydo would only apply to the rendered rows rather than the whole object (couldn't quick search on fields that aren't displayed).

If we had a more generic table component that took the records and a schema object, we could do this all in one place. I am open to expanding our scope to do that, but it's definitely going to be more than 60 lines of changed code.

leplatrem: Gotcha.

Then with the CSS nit, it would be good to ship 🙏

PR #3431, opened 2025-02-12
Fix #3263: Adding base64 file upload component

### LINKED ISSUE #3263: Add a new widget to store arbitrary files as base64 using a file input
This would allow users to store relatively small files as attributes on records instead of using attachments.

Typical use-case is to store the thumbnail of the file that is attached on the record.

In order to do this, we will need:
- a new widget that turns the file picked in the `<input type="file">` into base64
- to leverage this widget in [our Records form](https://github.com/Kinto/kinto-admin/blob/349a3748553324cd1ee14cf2a110895e68c189f7/src/components/record/RecordForm.tsx#L24-L43)

The record schema would be:

```yaml
properties:
  thumbnail:
    type: string
```
And the collection UI schema:

```
uiSchema:
  thumbnail:
    ui:widget: base64input
```

Then instead of contributing a new widget in the rjsf upstream project, we can follow the pattern that we currently have for our local JSON editor, and inject it ourselves during form instantiation:

https://github.com/Kinto/kinto-admin/blob/349a3748553324cd1ee14cf2a110895e68c189f7/src/components/JSONEditor.tsx#L12-L33

https://github.com/Kinto/kinto-admin/blob/349a3748553324cd1ee14cf2a110895e68c189f7/src/components/group/GroupForm.tsx#L44-L48

In this case we would replace occurrences of `base64input` by instances of our base64 widget.

See also:

- https://rjsf-team.github.io/react-jsonschema-form/docs/advanced-customization/custom-widgets-fields/

alexcottner:
Resolves #3263.
Adding a base64 encoding file upload option. See demo video below for how this prototype works.

This PR:

1. Adds a new RJSF widget to allow files to be "uploaded" as bse64 strings
2. Updates the Read-me with an example of this (should we update remote-settings docs too?)
3. Adds unit tests for the new component
4. Adds some validation to prevent too-large files from being uploaded (1MB cap)

See the demo video and screenshots below.

[demo video](https://github.com/user-attachments/assets/ad3610e7-a994-43fc-ab29-c55478f8b6af)

Validation error for too large files:
image


alexcottner:
@leplatrem and @nathan-barrett - How does this feel?

leplatrem:
Nice! 
micronit: I would maybe `file` in the widget name? `base64file`?

test/components/rjsf/Base64File_test.tsx
```diff
@@ -0,0 +1,99 @@
...
+import { fireEvent, render, screen, waitFor } from "@testing-library/react";
+import React from "react";
+
+const tinyTestImg =
+  "data:image/bmp;base64,Qk2aAAAAAAAAAIoAAAB8AAAAAgAAAAIAAAABABgAAAAAABAAAAAjLgAAIy4AAAAAAAAAAAAAAAD/AAD/AAD/AAAAAAAAAEJHUnMAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAACAAAAAAAAAAAAAAAAAAAA////AAAAAAAAAAD///8AAA==";
+const tinyTestTxt = "data:text/plain;base64,aGkK"; // "hi"
+const hugeTestTxt = `data:text/plain;base64,${"aGkK".repeat(1024 * 512)}`; // "hi" repeatedly
+
+class fileReaderMock {
+  error: FileReader["error"] = null;
```
alexcottner, line 11: This mock class may be worth committing to memory. This was the most difficult thing to develop.

nathan-barrett:
Awesome getting this in, lgtm

README.md
```diff
@@ -147,6 +148,37 @@ To run tests:
 ```
 $ npm run test-all
 ```
+
+## Custom RJSF fields
+We have expanded the RJSF functionality with a base64 encoded file upload. It can be used by declaring a string field on a collection's JSON schema and specifying the base64file widget in the UI schema. Examples:
```
leplatrem, line 153:
```suggestion
We have expanded the RJSF functionality with a base64 encoded file upload. It can be used by declaring a `string` field on a collection's JSON schema and specifying the `base64file` widget in the UI schema. Examples:
```

src/components/rjsf/Base64File.tsx
```diff
@@ -0,0 +1,91 @@
...
+          <Dash className="icon" />
+        </button>
+      )}
+    </div>
+  );
+}
+
+function CurrentValDisplay({ val }) {
+  if (!val) {
+    return <></>;
```
leplatrem, line 40: TIL :) 

Note that probably everywhere else we return `null`

src/components/rjsf/Base64File.tsx
```diff
@@ -0,0 +1,91 @@
...
+      ) : (
+        <Download className="iconLarge" title="Click to download" />
+      )}
+    </a>
+  );
+}
+
+async function getBase64Str(evt, changeCallback) {
+  const files = evt.target.files;
+  console.log("#### alex foo getBase64Str 1");
```
leplatrem, line 61: leftover

src/components/rjsf/Base64File.tsx
```diff
@@ -0,0 +1,91 @@
...
+          title="Image preview, click to download"
+        />
+      ) : (
+        <Download className="iconLarge" title="Click to download" />
+      )}
+    </a>
+  );
+}
+
+async function getBase64Str(evt, changeCallback) {
```
leplatrem, line 59: nit: it's called `onChange` elsewhere

src/components/rjsf/Base64File.tsx
```diff
@@ -0,0 +1,91 @@
...
+  const files = evt.target.files;
+  console.log("#### alex foo getBase64Str 1");
+  if (!files || !files.length) {
+    changeCallback(null);
+    return;
+  }
+
+  console.log("#### alex foo getBase64Str 2");
```
leplatrem, line 67: leftoever

src/components/rjsf/Base64File.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+  const readPromise = new Promise<string>((res, rej) => {
```
leplatrem, line 69: nit: I think `resolve` and `reject` are better names

test/components/rjsf/Base64File_test.tsx
```diff
@@ -0,0 +1,99 @@
...
+      errors = e;
+    };
+    render(<Base64File onChange={changeMock} />);
+    fireEvent.change(await screen.getByTestId("b64-file"), {
+      target: {
+        files: [tinyTestImg],
+      },
+    });
+    await waitFor(() => new Promise(resolve => setTimeout(resolve, 10))); // debounce wait
+    expect(val).toBeDefined();
```
leplatrem, line 79: Don't we want to assert that `val` is `"hi"`?

test/components/rjsf/Base64File_test.tsx
```diff
@@ -0,0 +1,99 @@
...
+    };
+    render(<Base64File onChange={changeMock} />);
+    fireEvent.change(await screen.getByTestId("b64-file"), {
+      target: {
+        files: [hugeTestTxt],
+      },
+    });
+    await waitFor(() => new Promise(resolve => setTimeout(resolve, 10))); // debounce wait
+    expect(val).toBeDefined();
+    expect(errors).toBeDefined();
```
leplatrem, line 97: Can we assert that the error is about size?

leplatrem (approved):
Nice! Thanks Alex for working on this.
 
I would recommend adding a test that does an assertion on the string field value that we obtain (eg `data:content/type;base64,`)

> [!IMPORTANT]
> We should probably add info about this in the records vs attachment section of our RS docs https://remote-settings.readthedocs.io/en/latest/getting-started.html#records-vs-attachments

leplatrem:
Renamed the PR title, because:

```
➜  kinto-admin git:(master) ✗ git log | grep -i 'fix' | wc -l
     296
➜  kinto-admin git:(master) ✗ git log | grep -i 'issue' | wc -l
      11
```

src/components/rjsf/Base64File.tsx
alexcottner: Yeah I think I was just avoiding that as we were getting a prop from rjsf called `onChange`. But I can just rename that one to `rjsfOnChange`.

leplatrem: then forget it 👌

test/components/rjsf/Base64File_test.tsx
alexcottner: Checking against our test values.

src/components/rjsf/Base64File.tsx
alexcottner: Nah, already did the thing

PR #3460, opened 2025-04-05
Issue-821 Replacing redux with hooks for servers

nitintecg:
For issue mentioned [here](https://github.com/mozilla/remote-settings/issues/821), Removing Redux for Servers and replacing with hooks .


src/sagas/session.ts
```diff
@@ -147,7 +147,7 @@ export function* setupSession(
     // Note, that "authenticated" can also mean "anonymous" if picked in the auth form.
     yield put(actions.setAuthenticated());
     // Store this valid server url in the history
-    yield put(serversActions.addServer(serverInfo.url, fullAuthType));
+    addServer(serverInfo.url, fullAuthType);
```
alexcottner, line 150: Hopefully we can remove the observable later once session is migrated.

src/hooks/servers.ts
```diff
@@ -0,0 +1,33 @@
+import { clearServers, loadServers, saveServers } from "@src/store/localStore";
+import { ServerEntry } from "@src/types";
+import { makeObservable } from "@src/utils";
+import { useEffect, useState } from "react";
+
+let state = makeObservable([]);
```
alexcottner, line 6: By defaulting to an empty array and trying to load later, we'll run into an a timing issue where we need to re-render the components. We can shortcut this by initializing state with what `loadServers` returns. Ex:

`let state = makeObservable(loadServers());`

Then we don't need line 12 either.

test/hooks/servers_test.ts
```diff
@@ -0,0 +1,43 @@
+import { addServer, clearServersHistory, useServers } from "@src/hooks/servers";
+import { renderHook } from "@testing-library/react";
+
+describe("servers hooks", () => {
+  describe("servers add", () => {
+    beforeEach(() => {
+      clearServersHistory();
+    });
+
+    it("should add server to recent history", async () => {
```
alexcottner, line 10: We are missing one test (that we were missing before :grimacing: ) that we properly return what is returned from `loadServers`.

nitintecg:
Thanks for review. I will make the changes by this week.

alexcottner (approved):
Excellent

PR #3467, opened 2025-04-11
WIP - replacing redux with hooks for records

alexcottner:
WIP - ignore for now

Replacing redux with hooks for actions associated with `records`. I'm expanding this scope a bit to include some things that used to be under `collection`.


alexcottner:
Canceling this for now, realizing I need to get routes adjusted first.

PR #3473, opened 2025-04-20
Issue-821 Replacing redux with hooks for heartbeat

nitintecg:


src/hooks/heartbeat.ts
```diff
@@ -0,0 +1,39 @@
+import { getClient } from "@src/client";
+import { HeartbeatState } from "@src/types";
+import { makeObservable } from "@src/utils";
+import { useEffect, useState } from "react";
+
+let state = makeObservable({
```
alexcottner, line 6: :thinking: Only one component needs to touch this hook, so I don't think we need to make it observable. But I see you're doing that so every time the queryHeartbeat function is called it will update state correctly.

I _think_ something like this would be cleaner and would also remove `checkHeartbeat` from the component:
```
port function useHeartbeat(): HeartbeatState {
  const [val, setVal] = useState<HeartbeatState>({
    success: true,
    response: {},
  });

  useEffect(() => {
    queryHeartbeat(setVal);
  }, []);

  return val;
}

export async function queryHeartbeat(setVal) {
  const client = getClient();
  try {
    const response: Record<string, any> = await client.execute({
      path: "/__heartbeat__",
      headers: undefined,
    });
    const success = !Object.values(response).some(value => value === false);
    setVal({
      success,
      response,
    });
  } catch (ex) {
    setVal.set({
      success: false,
      details: ex,
    });
  }
  setTimeout(() => {
    queryHeartbeat(setVal);
  }, 60000);
}
```

Note: I haven't fully tested this so I appologize if anything isn't right.

alexcottner (changes requested):
Removing 58 lines is great. I _think_ we can remove a couple more though. Have a small suggestion.

nitintecg:
Thanks for review. will look on this

nitintecg:
@alexcottner: can you review this ? 
thanks

alexcottner:
I am very sorry, I missed that you had replied. Adding a comment shortly.

I'd love for you to get this merged before I bring in this monster that I'm working through just so you're not dealing with the merge conflicts.

alexcottner (approved):
excellent work

PR #3588, opened 2025-07-24
Make sidebar toggable

leplatrem:
Users which land on RS with a direct link (nimbus) don't need the side bar

https://github.com/user-attachments/assets/4c4b3677-1844-473a-9db7-d6e44d3bbcae

TODO
* get review on hook / approach
* add tests


src/hooks/preferences.ts
```diff
@@ -0,0 +1,30 @@
+import { loadPreferences, savePreferences } from "@src/store/localStore";
+import { Preferences } from "@src/types";
+import { useCallback, useEffect, useState } from "react";
+
+const DEFAULT_PREFERENCES: Preferences = {
+  showSidebar: true,
+};
+
+export function usePreferences(): [
```
alexcottner, line 9: Excellent custom hook.

Currently this only affects 1 component. But it seems like we may be  looking to use this for other preferences in the future (like switching [useSimpleReview](https://github.com/Kinto/kinto-admin/blob/master/src/components/signoff/Review.tsx#L37) to it). This could cause an issue sharing state between multiple components if a user starts changing settings that could impact multiple components.

We could either:
1. Make this observable like [notifications]() so multiple components could set preferences and components would re-render as settings change.
2. Switch it to be like [useSimpleReview](https://github.com/Kinto/kinto-admin/blob/master/src/components/signoff/Review.tsx#L37) where it only cares about a single setting and will listen to storage change events.
3. Start using [context](https://react.dev/reference/react/useContext) now that redux is fully removed

I think I'd lean towards option 1 or 2. Option 3 would be good if we started implementing several settings (like theming).

test/components/Layout_test.tsx
```diff
@@ -102,4 +108,38 @@ describe("App component", () => {
...
+      renderWithRouter(<Layout />, routeProps);
+
+      expect(screen.queryByTestId("sidebar-panel")).toBeInTheDocument();
+      fireEvent.click(screen.getByTestId("sidebar-toggle"));
+      expect(setShowSidebar).toHaveBeenCalledWith(false);
+
+      // TODO/HELP?: why failing?
+      // await vi.waitFor(() => {
+      //   expect(screen.queryByTestId("sidebar-panel")).not.toBeInTheDocument();
+      // });
```
leplatrem, line 142: I don't get what's going on here 🦆

alexcottner: You've mocked the hook `set` function so you are now outside of the component lifecycle. So your component won't be re-rendering like normal.

But for this test, I don't think you need the re-render to happen. You just need to make sure the set event is called as expected when the button is clicked. You can test your hook behavior directly in a `test/hooks/preferences_test.ts` test file.

If you want to test a re-rendering with a different result, you could do something like this:
```
const setShowSidebar = vi.fn();
useShowSidebarMock.mockReturnValueOnce([true, setShowSidebar]);
useShowSidebarMock.mockReturnValueOnce([false, setShowSidebar]);

const rendered = renderWithRouter(<Layout />, routeProps);

expect(screen.queryByTestId("sidebar-panel")).toBeInTheDocument();
fireEvent.click(screen.getByTestId("sidebar-toggle"));
expect(setShowSidebar).toHaveBeenCalledWith(false);
rendered.rerender(<Layout />);

await vi.waitFor(() => {
  expect(screen.queryByTestId("sidebar-panel")).not.toBeInTheDocument();
});
```

I think this is unnecessary as it's a duplication of the 2 tests above. But that could be useful for you to know in the future.

alexcottner (approved):
:+1:

PR #3591, opened 2025-07-25
Scroll group list in permissions form

leplatrem:
Screenshot 2025-07-25 at 17 24 57

### DIFF
--- a/css/styles.css
+++ b/css/styles.css
@@ -451,7 +451,12 @@ form button[type="submit"].btn-primary {
 }
 
 .permissions-form .form-group.field-groups .container-fluid {
-  max-height: 600px;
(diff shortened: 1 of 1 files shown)


css/styles.css
```diff
@@ -449,6 +449,13 @@ form button[type="submit"].btn-primary {
   padding-left: 1em;
   margin-top: -60px; /* \o/ */
 }
+
+.permissions-form .form-group.field-groups .container-fluid {
+  max-height: 600px;
```
alexcottner, line 454: :thinking:   What if we made this a little more responsive like:
```css
  min-height: 300px;
  max-height: 50vh;
```

I see we're trying to make this more reasonable with as many groups as we have.

PR #3597, opened 2025-07-31
Ref #3140: stricter eslint

leplatrem:
Ref #3140

Remaining rules: 
* `@typescript-eslint/no-explicit-any` which has 73 occurences
* `@typescript-eslint/prefer-nullish-coalescing` which requires #2132 (`strictNullChecks`)
And `react-hooks/exhaustive-deps` will be done in a different PR


alexcottner (changes requested):
I think one coalesce needs to change back, but good other than that.

alexcottner (approved):

PR #3598, opened 2025-07-31
Enable react-hooks/exhaustive-deps

leplatrem:
* [ ] src/components/homepage/AuthForm.tsx
* [ ] src/hooks/collection.ts
* [ ] src/hooks/record.ts
* [ ] src/hooks/session.ts
* [ ] src/hooks/signoff.ts


alexcottner:
It's very easy to make an infinite re-render loop with `useEffect`. Sometimes an empty array or using a callback function to retrigger it is better than following the linter rules.

PR #3606, opened 2025-08-11
Use filters in all history tables

leplatrem:


src/components/bucket/BucketHistory.tsx
```diff
@@ -1,12 +1,27 @@
...
+  // Create filters object from current state
+  const filters = useMemo(
+    () => ({
+      show_signer_plugin: showSignerPlugin,
+      show_non_humans: showNonHumans,
+    }),
+    [showSignerPlugin, showNonHumans]
+  );
+
+  const history = useBucketHistory(bid, filters);
```
alexcottner, line 24: Feels like we could pass the `useXHistory` hook to the shared component from each parent component, then let the shared component handle everything instead of repeating the filters and hooks here.

alexcottner (changes requested):
Left a suggestion, I _think_ we can simplify this a bit.

src/components/bucket/BucketHistory.tsx
leplatrem: They don't have the same signature, so I am not sure how it would look.

But I should be able to pass an `onFiltersChange()` and avoid repeating the preference stuff

alexcottner (approved):

PR #3607, opened 2025-08-11
Fix mozilla/remote-settings#972: add collection compare view

leplatrem:
Fix mozilla/remote-settings#972

https://github.com/user-attachments/assets/23ac8031-5405-4aa3-a2cb-088b1a86d0d0


src/components/collection/CollectionCompare.tsx
```diff
@@ -0,0 +1,180 @@
...
+  const { bid, cid } = useParams();
+  const [loadingBuckets, setLoadingBuckets] = useState(true);
+  const [loadingRecords, setLoadingRecords] = useState(true);
+  const [loadingCollections, setLoadingCollections] = useState(true);
+  const [selectedBucket, setSelectedBucket] = useState("");
+  const [selectedCollection, setSelectedCollection] = useState("");
+  const hasAutoSelected = useRef(false);
+
+  const bucketsList = useBucketList();
+  useEffect(() => {
```
alexcottner, line 20: I don't think we need all these `useEffect` hooks here. We can do something simpler like:

```js
if (loadingBuckets && bucketsList) {
  setLoadingBuckets(false);
}
```

src/components/collection/CollectionCompare.tsx
```diff
@@ -0,0 +1,180 @@
...
+import CollectionTabs from "./CollectionTabs";
+import { useBucketList } from "@src/hooks/bucket";
+import { useCollectionList } from "@src/hooks/collection";
+import { useRecordList } from "@src/hooks/record";
+import React, { useEffect, useRef, useState } from "react";
+import { useParams } from "react-router";
+
+export function CollectionCompare() {
+  const { bid, cid } = useParams();
+  const [loadingBuckets, setLoadingBuckets] = useState(true);
```
alexcottner, line 12: I _think_ we can get a lot of these loading states based on their underlying lists being undefined. Right?

src/components/collection/CollectionCompare.tsx
```diff
@@ -0,0 +1,180 @@
...
+      hasAutoSelected.current = true;
+    }
+  }, [foundBucket, collectionsList, selectedCollection, cid, bid]);
+
+  // Fetch record lists for comparison
+  const leftRecords = useRecordList(bid, cid, "id");
+  const rightRecords = useRecordList(selectedBucket, selectedCollection, "id");
+
+  // Wait for both record lists to load
+  useEffect(() => {
```
alexcottner, line 67: `useRecordList` can tell us if we're loading or not based on it having a data array on it. It will reset to `{}` whenever the `bid`, `cid`, `sort`, or `cacheBust` values change.

PR #3650, opened 2025-08-27
Fixes #3649 - Replacing react-bs-notifier with react-bootstrap/Toast component

### LINKED ISSUE #3649: React 19 upgrade issue - toast notifications can cause a component crash
We're using `react-bs-notifier` for our toast notifications, which appears to not be compatible with react 19. We can also produce toasts with the `react-boostrap/Toast` components already. Let's switch to that instead.

alexcottner:
Fixes #3649 - replacing `react-bs-notifier` with `react-bootstrap/Toast` components.

Looking for feedback on this styling. This is at least readable (to me) and easy to understand when there is a warning/error. We can tune it further though.

image

### DIFF
--- a/src/components/Notifications.tsx
+++ b/src/components/Notifications.tsx
@@ -1,41 +1,64 @@
 import { removeNotification, useNotifications } from "@src/hooks/notifications";
 import type { Notifications } from "@src/types";
 import React from "react";
-import { AlertList } from "react-bs-notifier";
+import Toast from "react-bootstrap/Toast";
 
 function NotificationDetails({ details }: { details: string[] }) {
   if (details.length === 0) {
     return null;
   }
   if (details.length === 1) {
-    return <span>{details[0]}</span>;
+    return (
+      <Toast.Body>
+        <span>{details[0]}</span>
+      </Toast.Body>
+    );
   }
   return (
-    <ul>
-      {details.map((detail, index) => {
-        return <li key={index}>{detail}</li>;
-      })}
-    </ul>
+    <Toast.Body>
+      <ul>
+        {details.map((detail, index) => {
+          return <li key={index}>{detail}</li>;
+        })}
+      </ul>
+    </Toast.Body>
   );
 }
 
 export default function Notifications() {
   const notifications = useNotifications();
+
   const alerts = notifications.map(
-    ({ message: headline, details = [], ...attrs }, i) => {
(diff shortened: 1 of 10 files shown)


leplatrem:
👌 
What do you think of the opposite? Strong color on the title, and lighter on the toast body?

Screenshot 2025-08-28 at 09 58 47

```diff
diff --git a/css/styles.css b/css/styles.css
index cd87f4a1..78fa68ae 100644
--- a/css/styles.css
+++ b/css/styles.css
@@ -596,6 +596,14 @@ form button[type="submit"].btn-primary {

 .notifications .toast-body {
   font-size: 16px;
+  background-color: inherit;
+  filter: brightness(300%);
+  color: black;
+}
+
+.notifications .toast-header {
+  background-color: inherit;
+  color: rgba(255, 255, 255, 0.85);
 }

 .notifications .fade {
@@ -608,10 +616,6 @@ form button[type="submit"].btn-primary {
   right: .5em;
 }

-.notifications .bg-danger {
-  color: #fff;
-}
-
 .rjsf {
   margin-bottom: 1.25rem;
 }
```

src/components/Notifications.tsx
```diff
@@ -1,41 +1,64 @@
...
 
 export default function Notifications() {
   const notifications = useNotifications();
+
   const alerts = notifications.map(
-    ({ message: headline, details = [], ...attrs }, i) => {
-      const message = <NotificationDetails details={details} />;
-      return { id: i, headline, message, ...attrs };
+    ({ message: headline, details = [], timeout, type }, i) => {
+      console.log(details);
```
leplatrem, line 33: leftover

src/components/Notifications.tsx
```diff
@@ -1,41 +1,64 @@
...
+      return (
+        <Toast
+          className={`bg-${type}`}
+          key={`notification-${i}`}
+          autohide={timeout > 0}
+          delay={timeout > 0 ? timeout : undefined}
```
leplatrem, line 39: There could be a bug with this, not blocking, but if I click 6 times on the copy header in the top bar. Only 3 notifications disappear automatically

leplatrem (approved):
I think I prefer the opposite colors for the toasts, but not blocking my review on this

alexcottner:
I like the header colors being stronger. Commit that suggestion and fixed the lingering notification bug.

Issue #2047, opened 2021-09-16
Bootstrap 5 display issues


leplatrem:
# Generic

* Active tab has gray bottom line
![Screenshot from 2021-09-16 15-09-18](https://user-images.githubusercontent.com/546692/133617954-1497943e-9e4d-437e-b48e-10ad3c135bda.png)

* Add button icon is black (insconsistent with others (white))
![Screenshot from 2021-09-16 15-10-19](https://user-images.githubusercontent.com/546692/133618094-d187292d-280f-42d9-8bab-5e9a6ddee05f.png)

* Permissions overlap
![Screenshot from 2021-09-16 15-12-02](https://user-images.githubusercontent.com/546692/133618311-51246765-09b8-421f-897a-f9c806e1d256.png)

* Main app icon is invisible
![Screenshot from 2021-09-16 15-12-56](https://user-images.githubusercontent.com/546692/133618471-2c28b6bb-8d28-4a64-a023-35b2f0bbdd29.png)

* Toolbar icon is cropped
![Screenshot from 2021-09-16 15-13-34](https://user-images.githubusercontent.com/546692/133618581-8248b860-6cfa-47c3-91ed-bce48fc9dfb7.png)

* Loading table is all gray
![Screenshot from 2021-09-16 15-14-39](https://user-images.githubusercontent.com/546692/133618761-c84c0afd-5a07-4350-8a40-12948bf991ce.png)

# Notifications

* Close button is ugly:
![Screenshot from 2021-09-16 15-08-18](https://user-images.githubusercontent.com/546692/133617842-379d76f3-0495-48c9-b957-6fd8145075d2.png)

# Signoff plugin

![Screenshot from 2021-09-16 15-08-36](https://user-images.githubusercontent.com/546692/133617823-b75158ac-ca46-4854-86e1-869a0abe1451.png)

alexcottner:
Given that everything we're using is v4, and some of our dependencies don't have bootstrap v5 styling yet, I suggest we punt on trying to upgrade to v5 for now. Any objections?

alexcottner:
Closing this out for now as it seems high effort and low reward. Somebody re-open if this is a big issue. Will revisit when more dependencies are using v5.

leplatrem:
Makes sense, I'd leave the issue open but remove it from the milestone...

PR #3722, opened 2025-11-25
rjsf and bootstrap major upgrades

alexcottner:
Upgrading bootstrap to v5 and rjsf to v6. 

Tests pass and the site looks good to me running locally. Please do some manual testing to ensure I didn't miss any big styling issues.

Related to:
 - #2047 
 - #3699

### DIFF
--- a/src/components/collection/CollectionCompare.tsx
+++ b/src/components/collection/CollectionCompare.tsx
@@ -154,14 +154,12 @@ export function CollectionCompare() {
 
           {/* Bucket Select */}
           <div className="input-group mx-2" style={{ maxWidth: "220px" }}>
-            <div className="input-group-prepend">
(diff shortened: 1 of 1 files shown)


alexcottner:
> 1. I'm not a big fan of this blue

The colors for `btn-info` switched to that in v5. We can override it or switch to another variant though. See the [button variants](https://getbootstrap.com/docs/5.3/components/buttons/#variants).

> 2. Some margins are gone around the review steps

There appear to be a lot of subtle default margin/padding changes. White space for `container` and `card` have both been reduced by quite a bit. Both of which are used in this component. I added [spacing specifiers](https://getbootstrap.com/docs/5.0/utilities/spacing/) to adjust.

> 3. The popup close button is misplaced and ugly

Woops, missed that one. Caught it on the notification toast.

PR #3723, opened 2025-11-26
Fix mozilla/remote-settings#1086: add JSONEditor support in uiwdigets

leplatrem:
disclaimer: untested


src/components/BaseForm.tsx
```diff
@@ -10,6 +10,7 @@ import {
 } from "@rjsf/utils";
 import validator from "@rjsf/validator-ajv8";
 import React, { Component, ReactNode, useRef, useState } from "react";
+import JSONEditor from "./JSONEditor";
```
alexcottner: Prettier wants this moved to the top

PR #3875, opened 2026-05-29
Scroll panels instead of scrolling page

leplatrem:
This could help to address https://github.com/mozilla/remote-settings/issues/1278

And eventually we'll introduce https://datatables.net

https://github.com/user-attachments/assets/31a2a4d5-45d8-463e-a7d1-4d5fc7c03aa9

### DIFF
--- a/css/styles.css
+++ b/css/styles.css
@@ -8,6 +8,39 @@ html {
 html,
 body {
   background: #fafbfc;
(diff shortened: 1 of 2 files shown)


alexcottner (approved):
Feels good at the different screen sizes I played with (1200x800 and up).

PR #3886, opened 2026-06-09
fix: allow breadcrumbs to wrap on narrow screens

### LINKED ISSUE #3599: Should breadcrumbs wrap on narrow screens?
Image

remove334:
## Summary

- allows breadcrumbs to wrap on narrow screens
- keeps the styling scoped to the breadcrumb layout

Closes #3599.

## Test plan

- `git diff --check`
- `npm run cs-check` could not run because local dependencies are not installed.

### DIFF
--- a/src/components/Breadcrumbs/style.css
+++ b/src/components/Breadcrumbs/style.css
@@ -1,5 +1,25 @@
-.breadcrumbs {
+nav.breadcrumbs {
+  display: block;
+  overflow-wrap: anywhere;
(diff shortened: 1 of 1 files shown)


alexcottner:
Hello - Can you provide any before/after screenshots of what you're seeing it improve? Ex.

Current:
image

With this patch:
image

In my tests, it doesn't seem like it does much but remove some white space making it slightly more difficult to parse at times. But it's possible my test case isn't great.

